PR #265, opened 2011-09-20
Hitting Tab selects current highlighted option

### LINKED ISSUE #104: Selected Item Lost When Tabbing
If I scroll and / or autocomplete type to an item and hit tab, the selected item is lost and the combo box returns to "Select an Option".

adamjford:
While navigating through native selects using the keyboard, one can start typing the text of the desired option to select it, then hit the Tab key to move to the next element with the desired option remaining selected. This doesn't work with Chosen currently (the correct option is highlighted, but it is not selected once the Tab key is hit.)

Now the current highlighted option in a non-multiple Chosen select is selected when the Tab key is hit.

Resolves #104.


dannyw:
@pfiller is it possible to have this advance to the next field as well? per the original request: 

> While navigating through native selects using the keyboard, one can start typing the text of the desired option to select it, then **hit the Tab key to move to the next element** with the desired option remaining selected. 

Current behavior is that the tab key only selects it but keeps the focus on the current element.

pfiller:
As far as I can tell, there isn't a perfect solution to how Chosen handles tabbing. When it comes to issues of UX, Chosen normally defers to its frenemy, the <select> box. Normal selects don't allow _any_ tabbing when a dropdown is open. More importantly, tabbing doesn't ever _choose anything_.

I think there's a lot of practicality in that implementation decision. Making a choice to the extreme (no selection, no movement) prevents the weird UX edge cases that can crop up in these situations. It also makes sense in the most practical sense: the tab key is not for selecting things, its for moving things. Stopping all tab presses is a way to make sure the user's choice was deliberate.

That being said, I feel like Chosen has made a reasonable compromise in this regard. Tabbing selects the currently highlighted item, but it _shows_ you that it has done so by remaining focused. If you didn't expect the selection you can fix it easily enough. If it was intentional, you simply press tab again and move on.

I'd love to do some actual user testing and see how various versions play out. In the absence of evidence, I think I'd probably go further towards the "no tabbing" behavior than the "select and move on".

PR #424, opened 2012-01-03
Custom search separator

xcolour:
This patch makes the token used in search configurable and falls back on current token (" ") if not set.

In my project, I have a list filled with semi-hierarchical options that might look like:

```
app.cpu.load
app.mem.used
system.cpu.load
system.mem.used
```

And I'd like to be able to search for "cpu" and get the relevant entries.

This patch allows me to initialize a chosen field with

```js
foo.chosen({search_separator: "."});
```

And get the desired behavior.

The second commit also enables the use of regexes as the separator. I find this very useful for defining multiple characters to tokenize on (e.g. make three tokens out of "foo:bar-baz" with `/:|-/`).


pfiller:

Sorry this has sat open so long @xcolour. The hope is that by getting things under control, we can get to a place in the future where we can once again start adding new functionality (at which point we'd come back to closed PRs like this). Thanks for your patience and understanding!

PR #588, opened 2012-04-20
Updated file to use hide/show instead of setStyle('left', '-9000px')

egeste:
There was a point in time, long ago, where hiding elements by moving them far to the left was preferable for speed. But it's not really terribly relevant anymore

### DIFF
--- a/coffee/chosen.proto.coffee
+++ b/coffee/chosen.proto.coffee
@@ -16,8 +16,8 @@ class Chosen extends AbstractChosen
     super()
     
     # HTML Templates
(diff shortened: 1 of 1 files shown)


zhilevan:
I solve it with this method, 
first in chosen.css  comment all 9999 , then write this some line jquery to handle it
```javascript
  $(document).on('click','.chosen-single',function(){
    $('.chosen-drop').show();
  });
    $(document).on('click','.chosen-results li',function(){
    $('.chosen-drop').hide();
  });
```

PR #628, opened 2012-05-19
Support winnowing based on arbitrary undisplayed tags.

millimoose:
The checkin implements filtering based on the value of the attribute `data-tags` on the `<option>` element, if one is present.

For my app that uses Chosen, I needed to implement filtering the options (a list of people) based on their unaccented version. Instead of trying to strip the accents in Javascript, I figured a simpler, more flexible, and less invasive solution would be to implement filtering based on an arbitrary invisible alternate item key. The key would be generated on the server side as is appropriate for the locale / application, where the libraries for dealing with Unicode are better than in JS.

I tried to preserve the structure of the code as much as possible, the only major change is that the block of code responsible for determining whether an option matches the filter regex in `winnow_results()` is extracted into the methods `option_matches()` and `option_value_matches()` in the class `AbstractChosen`. Currently, when the match is against the tags, no highlighting of what caused the match is done in the UI.


onepiece2015:
+1

pfiller:

Sorry this has sat open so long @millimoose. The hope is that by getting things under control, we can get to a place in the future where we can once again start adding new functionality (at which point we'd come back to closed PRs like this). Thanks for your patience and understanding!

Issue #164, opened 2011-08-07
Programmatically open Chosen select options dropdown


wjcrowcroft:
Is there a way to programmatically open the Chosen select box? So, if a separate event is fired, to access a linked Chosen select and open the results?

Ideally would be something like this:

```
$('#someEl').click(function() {
    $('select#someId').trigger('chzn:open');
    // or
    $('select#someId').chosen("open");
});
```

If there's a way to do that by accessing internal Chosen functions, that works for now. Any thoughts? Thanks!

PR #771, opened 2012-08-17
Add two events for programatically activating / opening a field that has Chosen applied.

pfiller:
Addresses issues like #164

With this change, you can open or close Chosen by triggering events on the original form field.

``` javascript
$("#chosen_field").chosen().trigger("liszt:activate");
$("#chosen_field").chosen().trigger("liszt:open");
```

Note: triggering the open function on a field that has already been activated doesn't do anything.


hartleybrody:
For anyone else finding this via google, these triggered event names have changed in modern versions, and can be found in the [official documentation](https://harvesthq.github.io/chosen/options.html#triggerable-events).

```
$("#chosen_field").chosen().trigger("chosen:activate");
$("#chosen_field").chosen().trigger("chosen:open");
```

PR #839, opened 2012-09-30
Support for jQuery Validation

tombatron:
Chosen is awesome. jQuery validation is awesome. Put them together and it's as good or better than chocolate milk. 

Some notes:

Brand new to CoffeeScript. 

Brand new to Prototype. 

Both the Prototype and jQuery versions of Chosen work with the jQuery validation plugin.

This is my first pull request.


pfiller:
This has diverged pretty far from the current version of Chosen and it's not really anywhere near merge-able at this point 😢. I'm going to close un-merged.

There are definitely still issues with validation and Chosen, but they'll have to wait for another day. Thanks anyways, @tombatron!

PR #982, opened 2013-01-12
Allow user to deselect from single select chosen container from the keyboard when allow_single_deselect is true

MattCBowman:
If the chosen container has allow_single_deselect turned on and the user is operating from the keyboard, this fix will allow them to deselect without using the mouse. They simply select nothing and press return, which clears the currently selected item.


pfiller:
This PR is, unfortunately, way out of date and not easily merged. I'm going to close it 😢 

I think this is possibly a good approach for #2563 ... just let the user undo their selection by clicking on the thing they chose. I'll make a note there.

Thanks anyways, @MattCBowman. And sorry I wasn't very good about getting to review this in even close to a timely fashion.

Issue #493, opened 2012-02-10
Updating data-placeholder message when listz-updated


Abstremik:
Hello,

I have been using the chosen plugin with some success, however I have stumbled on a problem.

I have been trying to find a way to use this plugin in the following way: 

-> A chosen select is created with no options with data-placeholder set to "Empty"

``` html
<select id="choose" class="chzn-select" data-placeholder="Empty">  
<option></option>  
</select> 
```

``` javascript
<script type="text/javascript">  
$("#choose").chosen();  
</script>  
```

-> An event is triggered that retrieves data and builds the options to be selected resets data-placeholder and triggers listz-update

``` javascript
function showOptions(res){
   $('#choose').find('option').remove();            
   $("#choose").append($('<option></option>'));
   var obj = JSON.parse(res);
   for (i=0;i<obj.length;i++) {
             $("#choose").append($('<option></option>').val(obj[i]).html(obj[i]) );
   }
   $("#choose").attr("data-placeholder","Something");
   $("#choose").trigger("liszt:updated");
}    
```

However the message stays the same! What is the proper way ?

Issue #546, opened 2012-03-27
How to change the placeholder text on .change event 


rseshadri:
This code does not work

 $("#ctrycode").chosen().change(function () {
// $("#ctrycode").attr("data-placeholder", "+ 91");
 $("#ctrycode").data("placeholder", "+ 91").chosen();
});

Both these LOC would not work on chagne event. But the same code works on document.ready event.

Please help

PR #1107, opened 2013-03-25
Allow updating data-placeholder text with liszt:updated events

timcheadle:
I needed to update the data-placeholder text dynamically, and it turns out it was a pretty easy one-line change. Simply call `this.set_default_text()` whenever the `liszt:updated` event is triggered.

You can see an example of the update here:
http://jsfiddle.net/timcheadle/rLmCQ/

This should fix issues #493 and #546


tgice:
@timcheadle jsfiddle updated to work w/ new event name and something wrong with the original chosen include:

http://jsfiddle.net/rLmCQ/39/

To summarize, the two lines to run to update your placeholder text are:

```javascript
$('select').attr('data-placeholder', "Here's some new default text!");
$('select').trigger('chosen:updated');
```

Adding the comment since it took me longer than I would've liked to've figured this out (of course it helped to find this issue and jsfiddle).

Thanks for the improvement (5 years ago!)

PR #1161, opened 2013-04-24
Expose Option to Keep List Open After Selection

pfiller:
This is a new PR for #529. It handles the merge to get it up-to-date and is going to address some usability issues I found while testing that thread.


pfiller:
Oh jeez. This option was added at some point so this PR is no longer necessary. Closing!

Issue #1272, opened 2013-06-11
Bootstrap Modal


Sikwan:
Hey guys,

I was working on a project where chosen is used inside a bootstrap modal.

This is not working as expected but the fix is really quick, when a select is inside a div hidden the offsetwidth is equal to zero.

I think the best way could be to create the chosen object once the modal is shown but this is not very practical if this modal is called several time.

I got the correction in a fork, so tell me if you want a pull request. This is basically 3 lines checking if offsetwidth is 0 then using $().width() instead.

pierre-alain-b:
In my case, the behavior of Chosen was fine when using the "width intherit" option as follows:
<code>$(".chosen-select").chosen({width: "inherit"})</code>
Hope this helps.

pfiller:
Essentially a duplicate of #92. #2495 should solve it.

PR #1274, opened 2013-06-12
Bootstrap Modal correction

Sikwan:
As explained in this issue  : 

https://github.com/harvesthq/chosen/issues/1272

This is the correction I was talking about.


pierre-alain-b:
In my case, the behavior of Chosen was fine when using the "width intherit" option as follows:
<code>$(".chosen-select").chosen({width: "inherit"})</code>

PR #1282, opened 2013-06-17
Make sure Chosen isn't disabled before showing results.

### LINKED ISSUE #1235: Disabled multiselect chosen does not prevent the dropdown from opening and options being selected
With version 0.9.14 on chrome, firefox, ie, I'm disabling a multiselect chosen with:

``` javascript
$("#form_field_with_chosen").attr("disabled", true).trigger("liszt:updated");
```

Which works except that when you click on the field the dropdown opens and you can select options from it.

I've added a fiddle at:

http://jsfiddle.net/jalbrecht/CuQN3/1/

that shows the problem. In the fiddle:
1. Click on the disabled chosen field.
2. Note you can still select options.

Adding a disabled test to choices_click() seems to work but I didn't test keyboard navigation so I don't know if there are any issues there.

``` javascript
        AbstractChosen.prototype.choices_click = function(evt) {
            evt.preventDefault();
            if (!this.results_showing && !this.is_disabled) {
                return this.results_show();
            }
        };
```

pfiller:
@kenearley @stof this resolves #1235 

When a multiple select is disabled, Chosen currently lets you open up the results with a click (but offers no way to close). It's nasty, but comes with an easy fix.

Low-hanging fruit.


notionslate:
I recently upgraded to 1.2 webjar from 0.9.14 to get this fix but the problem appears to still be there. Can anyone confirm?

Issue #1306, opened 2013-06-26
Seperate positioning styling and fancy styling


lukeapage:
It would be great if chosen could 

a) split up the styling between that needed for positioning and that needed for styling
b) use a preprocessor like less and variables to allow people to tweak size variables that then refactor all of the css in a way that works
c) split up the css parts into different parts of the functionality so that a user can choose which parts of the styling to include.

doing this would be a good precurser to then writing a css adapter which could make chosen compatible with bootstrap etc.

pfiller:
Chosen's CSS is potentially getting a big overhaul in #2479. Feel free to take a look!

PR #1320, opened 2013-07-02
Generate CSS from SASS

koenpunt:
After a [previous PR](https://github.com/harvesthq/chosen/pull/1103) about using LESS for CSS @pfiller asked me if I was interested in doing something alike in SASS. 

I've put down a list of things that (probably) need to be done:
- [x] Replace manual vendor prefixes with mixins
- [x] <del>Split functional styling and visual styling</del>
- [x] Use shorthand color codes
- [x] Use unprefixed versions of properties (`border-radius`, `box-shadow`, `text-overflow`, `background-clip`)
- [x] Remove IE specific `filter:`

Bonus features:
- [ ] Make Chosen customizable using variables

If anyone has something to add to this list I'll be happy to hear so. As a follow up on #1306 I would like to hear your opinion also @lukeapage.

<del>**NB** I'm currently using Compass for mixins, but as Patrick mentioned in his comment on #1103:</del>

>  <del>we won't be able to easily use Compass</del>

<del>So maybe later the needed mixins can be included in the SCSS itself, without the need of Compass. </del>

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -47,8 +47,9 @@ module.exports = (grunt) ->
 
     compass:
       chosen_css:
-        files:
-          'public/chosen.css': 'sass/chosen.scss'
+        options:
+          specify:
+            ['sass/chosen.scss']
 
     cssmin:
       minified_chosen_css:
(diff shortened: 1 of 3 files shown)


cssguru:
What is the current status of the SASS version?  I'd like to vote up the include of a scss file in the download as well as the removal of the dependency on compass.  A bootstrap theme would be a nice addition as well.

PR #1329, opened 2013-07-04
Bootstrap Modal correction

funflin:
Fix for problem with bootstrap and modal window is the solution for #1272 of Sikwan but in coffeescript now in jQuery and Prototype

### DIFF
--- a/coffee/chosen.jquery.coffee
+++ b/coffee/chosen.jquery.coffee
@@ -22,6 +22,15 @@ class Chosen extends AbstractChosen
   finish_setup: ->
     @form_field_jq.addClass "chzn-done"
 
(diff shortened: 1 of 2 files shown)


pierre-alain-b:
I have the feeling that this PR was not merged within the tree (maybe I'm wrong). In any case, in my case, the behavior of Chosen was fixed when using the "width intherit" option as follows:
<code>$(".chosen-select").chosen({width: "inherit"})</code>

pfiller:
Thanks @funflin and sorry for taking 3 (THREE!) years to reply. Yowza. Going to close in favor of the newer (and simpler) #2495.

PR #1387, opened 2013-07-22
Feature: Default options

koenpunt:
I took a stab at making default options configurable.

For the jQuery version defaults can be set like:

```javascript
$.fn.chosen.defaults.allow_single_deselect = true;
$('.chzn-select').chosen();

// or

$.fn.chosen.defaults = {allow_single_deselect: true};
$('.chzn-select').chosen();

```

And for the Prototype version:

```javascript
Chosen.default_options.allow_single_deselect = true;
new Chosen($$('.chzn-select')[0]);

// or

Chosen.default_options = {allow_single_deselect: true};
new Chosen($$('.chzn-select')[0]);
```

I'm not quite sure about the merging, because in the jQuery version `$.fn.chosen.defaults` starts as a reference to `AbstractChosen.default_options`, but when `$.fn.chosen.defaults` gets overwritten the actual defaults are lost. So when instantiating I need to merge `AbstractChosen.default_options`, `$.fn.chosen.defaults` and `options` to get the correct options.

@harvesthq/chosen-developers: what are your thoughts on defaults?

### DIFF
--- a/coffee/chosen.jquery.coffee
+++ b/coffee/chosen.jquery.coffee
@@ -5,6 +5,7 @@ $.fn.extend({
     # Do no harm and return as soon as possible for unsupported browsers, namely IE6 and IE7
     # Continue on if running IE document type but in compatibility mode
     return this unless AbstractChosen.browser_is_supported()
+    options = $.extend({}, AbstractChosen.default_options, $.fn.chosen.defaults, options)
     this.each (input_field) ->
       $this = $ this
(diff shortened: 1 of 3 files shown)


pavdro:
Any progress on this one?
Really useful feature to set common config across the web app

pfiller:
@koenpunt and I just discussed this and we're going to close it. This is something we'd like to see added to Chosen at some point, but it's not a priority for now.

Issue #1233, opened 2013-05-23
Chosen with select-option on iOS


ghost:
Hello,

I think that I find a bug with chosen on iOS.

Without chosen, when you hit a select-option, options are shown instead of the keyboard, like this :
![y0l8e](https://f.cloud.github.com/assets/2360637/553414/83df7884-c384-11e2-98e7-73b29d580eba.jpg)

But with chosen, options are shown like in a desktop browser and alpha-numeric keyboard appear !

Can you solve this issue or give me a workaround ?

Thanks :)

Issue #1088, opened 2013-03-18
Needs work on mobile devices


thany:
I was thinking of using this plugin, because it just looks awesome. And it offers pretty much _exactly_ what I need...

However, it doesn't seem to have had mobile support in mind at all. It works shitty (excuse me) on a very small screen. When tapping the first demo, it pops up the onscreen keyboard _and_ it folds out the dropdown list. On such a small screen, it shouldn't do both. Most preferrably, it should just pop up the OS's default dropdown selector, because in the end, it's made to be most usable thing.

Tested on Android 4.2 stock browser.

Issue #382, opened 2011-12-06
Cannot scroll on iphone...


ghost:
Cannot scroll the results dropdown on the iphone... This makes this plugin a no go for future projects...

Issue #27, opened 2011-07-21
Mobile safari display is incorrect


danesparza:
The demo page displays incorrectly in mobile Safari on iOS.  Your supported browsers list Safari (and don't explicitly mention mobile Safari), but I would expect some type of fallback (to a regular select element, perhaps?)

PR #1388, opened 2013-07-22
Disable Chosen on iOS devices (iPad, iPod, iPhone)

koenpunt:
As a solution (until actual support is implemented?) for #27, #382, #1088 and #1233 I've flagged iOS browsers as unsupported.

### DIFF
--- a/coffee/lib/abstract-chosen.coffee
+++ b/coffee/lib/abstract-chosen.coffee
@@ -202,10 +202,10 @@ class AbstractChosen
   @browser_is_supported: ->
     if window.navigator.appName == "Microsoft Internet Explorer"
       return document.documentMode >= 8
-    if /iP(od|hone)/.test(window.navigator.userAgent)
+    if /iP(od|hone)/i.test(window.navigator.userAgent)
       return false
     if /Android/i.test(window.navigator.userAgent)
-      return false if /Mobile/.test(window.navigator.userAgent)
+      return false if /Mobile/i.test(window.navigator.userAgent)
     return true
 
   @default_multiple_text: "Select Some Options"


fritzgithub:
I agree with rockethouse (may 7 2014)
I use Chosen for design and UI but i do not at all need the search feature.
I cannot understand why this cannot be turned off for multiple selects (or do i keep overseeing things?).
Without search (and keyboard opening) on mobile devices, chosen would work like a charm (at least on my single test with firefox an android tablet).
Is there any chance i can deactivate search on multiple selects?

Mushr0000m:
Hi, any plans to make Chosen stable on iOS and Android ?

koenpeters:
If support for Chosen is eventually added I think it would be a good idea to simply use a full screen view (filter + list) for the phone and small tablet use cases. This mimics the native combo boxes and offers the best usability for users, especially a big chunk of the screen is taken up by a keyboard.

koenpunt:
@koenpeters I like that thought! But probably required a LOT of CSS updates.

koenpunt:
@mlettini maybe you can do a suggestion on what @koenpeters idea could look like?

mlettini:
@koenpunt @koenpeters That is a neat idea, and it might be doable, but it probably requires JS changes too. Because if we're talking CSS-only then you'd need to move the Chosen select out of it's position, which would require `position: fixed`. And if it's fixed, that usually doesn't work too well on mobile. And then the CSS would need to accommodate if the keyboard is open or not, and I'm not entirely sure how well that is handled (probably "not well at all"). And on top of all this, there's the different mobile OSes and their styles, and *then* we'd have to support certain mobile browsers... high effort IMO.

If I had the time, I might try a prototype, but alas, I don't right now.

koenpunt:
Issues with `position: fixed` was the first thing that came to mind for me as well, but if you take a look here: http://caniuse.com/#feat=css-fixed, you'll see that's probably not much of a problem anymore.

tjschuck:
@koenpeters @koenpunt @mlettini Any additional discussion about enabling Chosen on mobile should probably be a new issue.  You can link back to this thread, but this thread is "closed" (via merge), so it's more likely to get lost when people are looking for open feature issues.

@koenpeters Do you mind opening a new issue summarizing from your comment down?

Issue #1413, opened 2013-07-26
Use "top: -9999px" to hide elements instead of "left: -9999px" please!


neopersia:
Hi,

I love the chosen plugin but adding the "chzn-rtl" class to every select element in a right to left page is really ridiculous! It's even more annoying when it comes to multilingual pages, in which case you have to add the class pragmatically for every select element!

Why not using the "top" property instead? it's a lot simpler and it won't matter if the page is rtl or ltr. 
For the rest of the related css rules, adding a "rtl" class or even a standard dir="rtl" attribute to the body or html tag can render the need for "chzn-rtl" obsolete. That's the way it's handled in the dojotoolkit and I had no problems with that approach ever!

I could start writing codes right away but I've seen this argument in comments for some bugs here and I thought there might be a reason it's been ignored so far.

PR #1423, opened 2013-07-30
Optimize right to left compatibility

neopersia:
Fix for issue #1413


pfiller:
I don't think there's a problem here. #1159 moved the dropdown to the "hidden" side of the page based on RTL or not. Unless you're using a page with both text directions, that's an acceptable solution.

Thanks anyways, @neopersia

Issue #1386, opened 2013-07-22
Unintentional option selection when using a (Wacom) tablet as input device.


koenpunt:
At the office I'm using a Wacom tablet as primary input/mouse device.

Scrolling on such a tablet is done by holding a predefined button on the pen, but a release of this button while scrolling triggers an unintentional click (probably the `mouseup` event), and thus a option selection.

pfiller:
@koenpunt Is this still a thing? Do you really use a Wacom all day long? I want to know more!

koenpunt:
Defenitely, and at the office I use a Wacom all the time.

pfiller:
Crazy. It really does seem like a Wacom bug. No updates available? Does it affect you on other plugins/tools?

koenpunt:
I don't think it's a bug on Wacom's side, because there's actually a button going down and up, just not the primary button.

koenpunt:
Back at the office, just checked, there is actually a click of the primary button, but while holding a modifier button. And what I remember of when I opened this issue and possible solution ( #1431), that this modifier button isn't detectable in JS. But as it has almost been 3 years, I will reinvestigate this.

PR #1431, opened 2013-07-31
[WIP] Add result on mouseup only after a mousedown event was captured

koenpunt:
This is a fix for #1386.

When starting a scroll with a Wacom pen I have to press a secondary button and then drag the list, this doesn't trigger a `mousedown` event. But when I release to extend my scroll, the `mouseup` event is triggered, and thus the list is closed. Changing the result adding on click fixes this as a `click` event is only triggered when there has been a `mousedown` and `mouseup` event.

*WIP as there are still some issues in combination with mousedown/touchstart mouseup/touchend on iPad*


pfiller:
I'm a 👎 on this. If you press the mouse button on a Chosen single select before and then move the cursor down to the search results before releasing, you select a result. It reads complicated, but it's actually a simple (and imho useful) maneuver. I don't think supporting a specific device is a compelling reason to break that interaction.

Thanks anyways, @koenpunt.

Issue #1410, opened 2013-07-26
Using a chosen-element always fires a horizontal page scroll to the left


claudio10:
Didn't find this issue along the others already reported. It's not really a bug, but an unexpected behaviour. 

Summary of the issue:
When having a viewport that needs horizontal scroll bars, and a chosen-element is set rightoutside that viewport, the user will have to scroll the page to the right, until he can see the chosen-element he wants to use/select. Selecting it or selecting an li-item will result in a total scroll back left of the page, putting the just selected chosen-element outside of that viewport again.

Expected behaviour: The viewport should just keep on the selected element, without moving/scrolling.

Tracked OS: 
WinVista: FireFox 22.0, Chrome Version 28.0, Safari 5.1.7, IE 9.0
Android 4.1.2: Default Android Browser & Chrome (can't see where to get BrowserVers)

Steps required to reproduce the bug:

1. Go to http://harvesthq.github.io/chosen/

2. Minimize the browsers viewport until you see the page with bottom scroll bars like this:
![image](https://f.cloud.github.com/assets/5096188/862117/e6148146-f5e5-11e2-84de-1eece0569d04.png)

3. Scroll to the right until you can see the right chosen-element like this:
![image](https://f.cloud.github.com/assets/5096188/862122/2edaf2ac-f5e6-11e2-8424-a2f58c09d4e6.png)

4. Select a option of that second chosen-element and you will see the page being scrolled totally back to the left, putting the just selected element outside the viewport again.

The behaviour will occure whenever the page has to be scrolled right for a chosen-selection, e.g. wide-screen pages with multiple side by side chosen-elements, mobile devices/tablets with reduced viewports. 

Best wishes and thx for your great solution & contribution!!!

Claudio

Mike0609:
Fix by css
34 line - .chosen-container .chosen-drop
change 

left: -9999px;
to
left: 0;

and add display: none;

46 line - .chosen-container.chosen-with-drop .chosen-drop
replace: 

left: 0;
to
display: block;

and everything works fine

qwelyt:
@Mike0609 That solution does not seem to work. Any select that does not have a searchbar produces the left-bug.

Mike0609:
https://github.com/qwelyt I'm sorry, I checked and found inaccuracies, here is a link to a working example: http://chirskiy.ru/chosen_fix/

qwelyt:
@Mike0609 As far as our product goes, that works wonders! Thanks!
[...]

PR #1436, opened 2013-08-01
Fix for #1410: Do not set input focus if search is disabled (i.e. input ...

gorner:
...is offscreen)


pfiller:
I think we can solve this problem with CSS (a la something like #1990). We may also want to investigate clipping as suggested by @koenpunt (http://a11yproject.com/posts/how-to-hide-content/).

Thanks anyways, @gorner.

PR #1439, opened 2013-08-01
Change the Way Width is Determined

pfiller:
@harvesthq/chosen-developers 

This fixes some issues that come when applying Chosen to a `select` that is not visible (when you haven't set `width: n` for whatever reason). Basically, it uses jQuery & Prototypes magic methods for getting width if they're set and falls back to `offsetWidth` if they're not. As always, a width specified in the option object will take precedence.

I went a little crazy testing this. Each of these fiddles demonstrates 4 ways you would apply width to Chosen in both a container that is hidden on page load and a container that is visible on page load. As you can see, with the exception of a hidden select that has no style applied to it, the new version does a wonderful job of matching size:

- [jQuery 1.0 version](http://jsfiddle.net/tTn5a/2/)
- [jQuery fix-hidden-width version](http://jsfiddle.net/geJ7t/1/) 
- [Prototype 1.0 version](http://jsfiddle.net/JUFUE/2/)
- [Prototype fix-hidden-width version](http://jsfiddle.net/gsavJ/1/) 

**Please Note:** This doesn't fix the % width issue (people will still need to specify those via `width: %`. I only mention this because one of these examples uses a percentage width set via a class. If you adjust the width of the output area, Chosen might not match it's original field. The best thing to do is adjust the width of the area and then click "Run" again. (Can you tell I learned this the hard way?)

### DIFF
--- a/coffee/chosen.jquery.coffee
+++ b/coffee/chosen.jquery.coffee
@@ -497,7 +497,7 @@ class Chosen extends AbstractChosen
 
       f_width = @container.outerWidth()
 
-      if( w > f_width - 10 )
+      if( f_width && w > f_width - 10 )
         w = f_width - 10
 
       @search_field.css({'width': w + 'px'})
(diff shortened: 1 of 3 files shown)


pfiller:
Going to close this olllllllld PR. @koenpunt's #2495 is probably a better solution ultimately.

PR #1442, opened 2013-08-01
Prevent chosen-drop elem from appearing when empty

alademann:
Without checking to see if the total number of chosen results matches the total number of results in the `@results_data` array, when the `results_show` method is called, the chosen-drop elem will appear as an empty white box with a shadow - which to me, is an undesired effect.

The code changes are pretty straightforward.  I didn't see anywhere to write unit tests - although I'd be happy to write one for this if that is needed. 

Thanks for considering the changes!

### DIFF
--- a/coffee/chosen.jquery.coffee
+++ b/coffee/chosen.jquery.coffee
@@ -176,7 +176,10 @@ class Chosen extends AbstractChosen
 
     @results_data = SelectParser.select_to_array @form_field
     @results_count = @results_data.length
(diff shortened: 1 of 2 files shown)


pfiller:
This PR was from a time when Chosen actually removed items from the list after they had been selected. It no longer does so and so selecting all the options looks ok (to me):

not too shabby

Thanks anyways, @alademann. It's 3 years later, but I still appreciate your efforts.

Issue #1463, opened 2013-08-09
wrong letters underlining


yurks:
Given:
-- single selectbox with firstnames and lastnames
and
-- .chosen() applied with default values (so, "search_contains" is false by default)
When
-- selectbox is open
and
-- start typing leters in searchbox
and 
-- both words contains typed letters
Then
-- wrong word is underlining

![image](https://f.cloud.github.com/assets/4581398/938753/f0510556-0104-11e3-9ce4-fc0dfbfc706e.png)

koenpunt:
Fixed by https://github.com/harvesthq/chosen/pull/2692

PR #1483, opened 2013-08-16
Use word boundary matching

### LINKED ISSUE #75: Search bug
One more problem...
http://www.tophouse.ru/price/index-new.php
I have "Top Stone" in my list. If I enter "top " I see 1 result, but If I enter "top s" I not see anything...
Why?

pfiller:
@harvesthq/chosen-developers 

This is a new PR for @koenpunt's excellent #1465. In his words:

> By making use of word boundary matching (http://www.regular-expressions.info/wordboundaries.html) the split word search is much cleaner (no more actual splitting of strings), and the search results are better.

I want to be very careful about merging this in, but I'm loving what I see so far. With default options set:

<table>
<tr>
  <th>Search String</th>
  <th>Expected to Match</th>
  <th>Old way</th>
  <th>New way</th>
</tr>
<tr>
  <td>Democratic Republic of</td>
  <td>Congo, The Democratic Republic of The</td>
  <td>:poop:</td>
  <td>:+1:</td>
</tr>
<tr>
  <td>of America</td>
  <td>United States (of America)</td>
  <td>:poop:</td>
  <td>:+1:</td>
</tr>
<tr>
  <td>Holland</td>
  <td>Netherlands/Holland</td>
  <td>:poop:</td>
  <td>:+1:</td>
</tr>
</table>

Here's what I think we should do before this gets a final :shipit:
- [ ] Add unit tests for verifying that search results are correct. To do this, we might need to break the actual search match function out of the winnow method.
- [ ] Performance Test with 1000s of items (perhaps replacing test + match with a single match call)
- [ ] Determine if we need to make an API change related to `enable_split_word_search` or `search_contains`. I don't think you should be able to turn off `enable_split_word_search` if `search_contains` is `true`, but I want to do some experimenting there.

Fixes #75 and #1463

### DIFF
--- a/coffee/lib/abstract-chosen.coffee
+++ b/coffee/lib/abstract-chosen.coffee
@@ -214,6 +214,7 @@ class AbstractChosen
 
   get_search_regex: (escaped_search_string) ->
     regex_string = if @search_contains then escaped_search_string else "\\b#{escaped_search_string}\\w*\\b"
+    regex_string = "^#{regex_string}" unless @enable_split_word_search or @search_contains
     new RegExp(regex_string, 'i')
 
   choices_count: ->
--- a/coffee/lib/abstract-chosen.coffee
+++ b/coffee/lib/abstract-chosen.coffee
@@ -167,7 +167,6 @@ class AbstractChosen
     searchText = this.get_search_text()
     escapedSearchText = searchText.replace(/[-[\]{}()*+?.,\\^$|#\s]/g, "\\$&")
     regex = this.get_search_regex(escapedSearchText)
-    highlightRegex = this.get_highlight_regex(escapedSearchText)
 
     for option in @results_data
 
@@ -188,12 +187,14 @@ class AbstractChosen
         option.search_text = if option.group then option.label else option.html
 
         unless option.group and not @group_search
-          option.search_match = this.search_string_match(option.search_text, regex)
+          option.search_match = regex.test(option.search_text)
+
           results += 1 if option.search_match and not option.group
 
           if option.search_match
             if searchText.length
-              startpos = option.search_text.search highlightRegex
+              match = regex.exec(option.search_text)
+              startpos = match.index
               text = option.search_text.substr(0, startpos + searchText.length) + '</em>' + option.search_text.substr(startpos + searchText.length)
               option.search_text = text.substr(0, startpos) + '<em>' + text.substr(startpos)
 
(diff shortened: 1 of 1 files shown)


oniofchaos:
This would be helpful for me as well.

chrisvfritz:
I think after 2 years it's pretty clear that this relatively simple fix is not going to happen. I recommend using [selectize](http://brianreavis.github.io/selectize.js/) instead. You not only won't deal with this particular issue, but in my experience, it's been better in every way.

alxndrmlr:
Agreed. I went with [select2](https://select2.github.io/)

koenpunt:
@pfiller should we add the target label to this one for our next hangout?

tjschuck:
@koenpunt Is there anything actually blocking this (aside from needing to be rebased)?  Are @pfiller's three todos on the main body of this "real" todos?  Which ones are blockers vs. nice-to-haves?

koenpunt:
Rebase is done, and no real blockers I think;
- testing could have some impact.. but since we're going with almost no tests currently, it's hard to say what we have to test in the first place.
- Second TODO about performance; I don't expect any issues there, especially since we're going from 2 regular expressions to just 1.
- Also I don't think the public API changes (and I believe we came to an understanding on that [about 4 years ago](https://github.com/harvesthq/chosen/pull/1483#issuecomment-22818111) already)

koenpunt:
Also would like to quote @pfiller 

> Wow, this is great. I've been testing it and testing it and I think it works great. That being said, it's a pretty significant change to the way searches are performed in Chosen so I want to tread lightly.
> @pfiller commented on [16 Aug 2013](https://github.com/harvesthq/chosen/pull/1465#issuecomment-22796481)

Is 4 years light enough? 🤓

tjschuck:
@koenpunt I agree.

* Lack of tests has never stopped us before 😏 
* We already have purported very-long-list performance issues, and I think this will make the happy path much happier.
* I also don't see any public API change.

If you're giving a 👍 to the code, I say ship it!

(In a rare fit of proximity, I can physically see @pfiller _right now_.  I'll try to force him to... do anything? Respond? ¯\\\_(ツ)_/¯  Anyway, don't wait for him if you feel good about it.)

koenpunt:
It goes a bit against our (or mine) prior notice;

> From this version on, until further notice, no new features will be added, so pull requests for features will not be accepted. However, pull requests for bugfixes are still very much appreciated.

But technically it's not a _new_ feature anymore.

koenpunt:
Just had a final thought that maybe it would be a downside that searching for `(` or `)` wouldn't work anymore, but apparently that didn't work anyway 😂 

image

tjschuck:
It's also not really a new feature — it's just fixing an existing feature to work way better.  I might even call it a bugfix, based on all the linked issues!

@pfiller said @satchmorun might have some insight here, since he was allegedly poking around similar areas of Chosen recently.  Regardless, I think we should merge this by Monday at the latest.  If you want to wait and see if @pfiller or @satchmorun has anything to say, go for it, but if you don't, also go for it.

koenpunt:
Done 🚀

tjschuck:
@koenpunt :heart: :heart: :heart_eyes_cat: :heart:

PR #1510, opened 2013-08-26
Right to Left initialization option

kenearley:
@harvesthq/chosen-developers I've pulled in [@koenpunt's branch](#1491) and updated the docs to note deprecation of the classname approach.


koenpunt:
Why did this linger for so long?

PR #1537, opened 2013-09-09
Ignore invisible soft hyphen characters when matching options

antris:
Conditionally invisible soft hyphens (http://en.wikipedia.org/wiki/Soft_hyphen) can be used to break long words to many rows so the UI stays clean. There is no reason to take soft hyphens into account when matching text because users never enter it in the search box. It only introduces a bug where a long word cannot be matched because the user has not entered the invisible character in the search field.

### DIFF
--- a/coffee/lib/abstract-chosen.coffee
+++ b/coffee/lib/abstract-chosen.coffee
@@ -153,7 +153,7 @@ class AbstractChosen
                 
         unless option.group and not @group_search
 
-          option.search_text = if option.group then option.label else option.html
+          option.search_text = (if option.group then option.label else option.html).replace(/\u00AD/g, '')
           option.search_match = this.search_string_match(option.search_text, regex)
           results += 1 if option.search_match and not option.group
 


pfiller:
I don't think there's much need to merge something like this into Chosen as it seems like an edge case. If we were going to start replacing characters before search (I'm not sure there's demand for this), I'd prefer it be configurable so we don't have to be opinionated with our defaults.

Thanks anyways, @antris.

antris:
I cannot fathom someone having the opinion that soft hyphens should be searchable. Is it really a matter of opinion?

pfiller:
It's not that I disagree with you. It's that we're not going to add code that does something that few people seem to be asking for. Every line of code adds a maintenance burden and, as you can tell from the large number of (old) open pull requests, we're not doing a very good job of fighting that burden.

Luckily, the way open source code works is that you can continue to use your fork that does the thing you need. Even better, because of GitHub, other people who might want the same functionality have a way to find your changes. Isn't that cool?

Issue #135, opened 2011-08-02
Multiselect with shift key


brechtvhb:
The multi select should allow us to select more then items at once by holding the shift key, like the regular non enhanced version does.

casph:
+1

patawa91:
I'm working on a pull request right now that allows user shift click and highlight multiple. Then the second click adds the highlighted items. Has a pull request for this type of thing been submitted? Is there even a possibility of this functionality being pulled in?

kevin-brown:
> Has a pull request for this type of thing been submitted?

Yup, there's one at https://github.com/harvesthq/chosen/pull/1563

laantorchaweb:
be aware, my PR for adding multiselection holding shift key is from 2013, I had rebased the code and updated it last week so it's ok to merge but I don't think that this will happen any time soon. This issue is from 2011...

NeumannPSG:
+1

NeumannPSG:
+1 for this feature

cahrens:
+1 also for this

PR #1563, opened 2013-09-18
#135 added holding shift functionality for multiselection of several items

laantorchaweb:
Added functionality for selecting multiple items holding the shift key, the same behavior that is expected in plain html multi select. It work in both direction, up to bottom and reverse. Tested on ie8, ie9, ie10, firefox and chrome.


qwelyt:
@laantorchaweb We need this in our product. Unfortunately, your PR is out of date. Would you be able to rebase your code so it can be merged with Chosen?

laantorchaweb:
Sure, I will update the repo and see if the previous solution works fine or I'll implement something else.

laantorchaweb:
The rebasing seems to have worked fine, I've build the project and both jquery and prototype versions works fine.  Let me know if you run into any issues.

qwelyt:
@tjschuck How about that merge, ey?

qwelyt:
@laantorchaweb Seems to work OK. There is however a problem when you select things from different optgroups. Going from an option without an optgroup to an option in an optgroup produces some quite odd behaviour, like only selecting the things outside the optgroup, and keeping the selected position in the optgorup resulting in the rest being selected when you open it up again.
There also is a strange behaviour of the same result being selected multiple times, somehow..

But it's an improvement from before, shift select actually works!
The highlighted options might also need some looking into. Right now I get this: I press shift+left moust button, the first option is selected. I hold shift and move my mouse down and I expect options to be marked as selected as is the case in the normal multiselect. They do however not become selected. But when I press the left mouse button again, they all become selected. So it works, you just don't see that it works.

Gemfile.lock
```diff
@@ -11,6 +11,7 @@ GEM
 
 PLATFORMS
   ruby
+  x86-mingw32
```
tjschuck: :scissors: Please remove this line.

coffee/lib/abstract-chosen.coffee
```diff
@@ -162,7 +166,7 @@ class AbstractChosen
           results_group = @results_data[option.group_array_index]
           results += 1 if results_group.active_options is 0 and results_group.search_match
           results_group.active_options += 1
-                
+
```
tjschuck: :scissors: Please remove this and all other whitespace-only changes from the PR.

laantorchaweb:
I've updated the code to be up to date with master, about the groups it's true that you can't multi select items when they are on different groups, I'll see if I can make this happen and also add classes to the items that are being hovered while holding shift.

ComeBurguburu:
+1

pfiller:
So, this is the kind of thing that I would normally be in favor of. But ... it adds a pretty large amount of code for something only advanced users are likely to use. Your average web user doesn't know they can hold shift down in a normal select and I doubt they would know to do it here.

I know a lot of people want this feature, but I think the maintenance burden is likely to be too much. Luckily, there's a fork linked right here that those people can use!

Thanks anyways, @laantorchaweb. Sorry it took THREE YEARS to make a decision...

PR #1610, opened 2013-10-15
Feature: Option to show group labels in selected option(s)

koenpunt:
I've rebased the work of @rtlong's PR #567 on current master.

One thing I've changed is that in both single and multiple select the separator between the group name and option name is a colon (`:`), instead of an arrow symbol and a colon.

### DIFF
--- a/public/options.html
+++ b/public/options.html
@@ -109,6 +109,13 @@
             <p><strong>Note:</strong> this is for multiple selects only. In single selects, the selected result will always be displayed.</p>
           </td>
         </tr>
+        <tr>
+          <td>include_group_label_in_selected</td>
+          <td>false</td>
+          <td>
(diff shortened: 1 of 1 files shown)


andreialecu:
Not sure if this is the right place to report this, but when you trigger "chosen:updated" after setting the selected item on the main select element, the label isn't updated properly to include the option group name. It only works when clicked, but not when programatically set.

tjschuck:
@andreialecu Thanks for the report.  Can you open up a [new issue](https://github.com/harvesthq/chosen/issues/new) for that?  Please reference this PR when you do.

If at all possible, can you also please link to a publicly accessible page (or a [Fiddle](http://jsfiddle.net)) that shows the issue so we can attempt to reproduce?

Thanks!

andreialecu:
I actually fixed, it I'll send a PR.

andreialecu:
PR https://github.com/harvesthq/chosen/pull/2217

PR #1617, opened 2013-10-18
Adds a semi-colon that was missing from the set up.

mlissner:
Super minor. Just a semi-colon.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -1414,7 +1414,7 @@
       <p>Using Chosen is easy as can be.</p>
       <ol>
         <li><a href="https://github.com/harvesthq/chosen/releases">Download</a> the plugin and copy the chosen files to your app.</li>
-        <li>Activate the plugin on the select boxes of your choice: <code class="language-javascript">$(".chosen-select").chosen()</code></li>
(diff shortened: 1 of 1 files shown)


arunabhp75:
I don't get the ideas of triage, can someone help me out?

koenpunt:
Thanks, but semicolons are optional in JavaScript

koenpunt:
@arunabhp75 triage is about assisting or providing an outside look on issues, which could help in solving those. In this case this PR was probably overlooked, and you being directed here to triage brought it under the attention again. 👍🏼

mlissner:
@koenpunt semi-colons are *sometimes* optional. The reason that I filed this fix is because when I copy/pasted the code from the documentation, my editor threw an error. Seems better to default to having it, so that copy/paste works properly?

I mean...do as you please, it's just a semi-colon, but it's the kind of thing that can trip up a JS newbie.

koenpunt:
@mlissner the opposite is true; semicolons are _sometimes_ required. A nice writeup by @mislav here: http://mislav.net/2010/05/semicolons/

PR #1626, opened 2013-10-25
Add 'chosen:search_results_updated' event

juanghurtado:
This event will be triggered on every update of the search results
area (via `update_results_content()`). It can be used, for example,
in case you need to attach extra info to your search result items.


juanghurtado:
Closing, too old

PR #1630, opened 2013-10-29
SearchChoice pill: inherit classes on the option

sdemjanenko:
This lets classes attached to the option DOM be passed through to the selected pill that is created.  This would be super useful for styling pills.


blazecc:
+1

PR #1646, opened 2013-11-07
Use style width if select list is hidden and width option not set.

frogmonster:
If the select list or its container is hidden, width is set to 0px. I know that we can set the width option on the chosen() method (ie, chosen({width: '100px'}), but that forces us to use a chosen() statement for each and every hidden select vs being able to simply call $('.chosen-select').chosen() (and being able to control look and feel with page level css).

If there is no width option and offsetWidth is 0px, then chosen should try to use the width style attribute on the select list.

Excuse any erroneous coffeescript, as this is my first time writing in it :-)

UPDATE: I just realized that my pull request could be more efficient by caching the offsetWidth property (which appears to cause a dom reflow everytime it's called). When I get a sec, I'll update the coffeescript...

### DIFF
--- a/coffee/lib/abstract-chosen.coffee
+++ b/coffee/lib/abstract-chosen.coffee
@@ -223,7 +223,13 @@ class AbstractChosen
       else this.results_search()
 
   container_width: ->
-    return if @options.width? then @options.width else "#{@form_field.offsetWidth}px"
+    return if @options.width? 
+      @options.width 
+    else if @form_field.offsetWidth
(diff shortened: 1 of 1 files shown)


pfiller:
Thanks for opening this @frogmonster. I'm sorry it's been open for a year and a half with no comment from a maintainer. Geez.

I think this approach is solid, but I think @koenpunt's #2495 from this year is probably a little more solid. Gonna close in favor of that. Thanks again!

PR #166, opened 2012-08-27
Feature: Option adding (with custom callback)

koenpunt:
### Option adding

You can specify a `create_option` option, which can be a boolean or callback function.

Example for allowing adding of new items:

``` javascript
$(function() {
  $(".chzn-select").chosen({
    create_option: true,
    // persistent_create_option decides if you can add any term, even if part
    // of the term is also found, or only unique, not overlapping terms
    persistent_create_option: true,
    // with the skip_no_results option you can disable the 'No results match..' 
    // message, which is somewhat redundant when option adding is enabled
    skip_no_results: true
  });
});
```

Or with a callback and an AJAX request to, for example, save the terms in your database:

``` javascript
$(function() {
  $(".chzn-select").chosen({
    create_option: function(term){
      var chosen = this;
      $.post('add_term.php', {term: term}, function(data){
        chosen.append_option({
          value: 'value-' + data.term,
          text: data.term
        });
      });
    }
  });
});
```
### Custom text

Through the options it is also possible to customize the text used in the 'Add option' link, like so:

``` javascript
$(function() {
  $(".chzn-select").chosen({
    create_option_text: 'Create category'
  });
});
```

PR #1647, opened 2013-11-08
[WIP] Make Chosen more extendable

koenpunt:
After keeping my PR for option adding (#166) up to date for over 2 years I got a suggestion from @mculp: koenpunt/chosen#38.

And by extracting more code into separate methods extending Chosen should become a breeze.


pfiller:
In an effort to clean up Chosen's PRs once and for all, @koenpunt and I just decided to close this one. Making Chosen more extendable would solve a lot of the issues people have with it, but this change is difficult to make without breaking most of the open PRs we're looking at.

We're not saying no forever, but we are saying no for now. As Koen says, "for a better future".

PR #1667, opened 2013-11-25
allow the no_results_text param to be a function

stryju:
this comes in handy when, for example, you want to transform the 'no results' html content

like `'the term "foo" was not found - click [here] for more info'` where `[here]` is a link containing the term(s)


stryju:
ಠ_ಠ

Issue #1655, opened 2013-11-15
jspm support


guybedford:
I've recently added this to jspm:

http://jsbin.com/EfOYaBE/2/edit

The "github" endpoint in jspm (`github:harvesthq/chosen`) reads the releases feature from Github and uses the code directly from there.

In order for the code to play with with jspm, I needed to add the following package.json override in jspm:

``` javascript
{
  "main": "chosen.jquery.js",
  "files": ["chosen.jquery.js", "chosen.css", "chosen-sprite.png", "chosen-sprite@2x.png"],
  "shim": {
      "chosen.jquery": ["./chosen.css!", "jquery@2.0"]
  },
  "buildConfig": { "uglify": true }
}
```

This allows the modular shim config, and minification with source maps to work at the jspm Github CDN. Allowing for the modular loading.

If the above were included in a `package.json` file in the release itself, this would add jspm support by default (instead of us needing to override each time).

If there is any chance of this being included, it would be a great help.

PR #1669, opened 2013-11-26
add jspm package.json config

guybedford:
Further to #1655, I've included the necessary changes in the package.json here.

I've relaxed the jQuery constraint, so that this can be used with any version of jquery as simply as:

```javascript
  jspm.config({
    map: {
      jquery: 'jquery@1.8'
    }
  });
```

Being included before the require to `jspm.import('chosen')`

This allows the exact version to be environment-specific. Semver ranges will be supported in a future version as well, but that is a few months away.

The demo is still available here - http://jsbin.com/EfOYaBE/2/edit

Any further questions just ask.

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,5 +70,13 @@
     "dependencies": {
       "jquery": ">=1.4.4"
     }
+  },
+  "jspm": {
+    "main": "chosen.jquery.js",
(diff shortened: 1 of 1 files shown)


pfiller:
I think we should close this as there doesn't seem to be much clamoring for JSPM. We did finally agree to support Bower and NPM, but it's been a total hassle. I don't want to add anything else unless there's a super compelling reason to do so.

Thanks anyways, @guybedford.

PR #1675, opened 2013-12-03
Fixed result-selected class being added to more than one option after the first use of the drop down.

nodesman:
The result-selected class can be sometimes used to hide the currently selected option to avoid redundancy in the options being shown in single mode. But because of the issue that this change fixes it is sometimes added to two options - the current and the previously selected one.


koenpunt:
This have been around for almost 3 years now, without followup. If you still think this is an problem, [open a new issue](https://github.com/harvesthq/chosen/issues/new)

PR #1704, opened 2013-12-27
Added title attribute to options and optgroups

### LINKED ISSUE #1692: Add title attribute (tooltip) to Chosen option tags
http://stackoverflow.com/questions/14569903/how-to-add-title-attribute-to-chosen-option-tags

The instructions from that stackoverflow thread works just fine for me. Why cannot this be standard behavior of chosen?

/Anders

metaist:
Minor patch to close #1692. The `title` attribute from the original `<option>` and `<optgroup>` are preserved when generating the `<li>` for chosen.

### DIFF
--- a/coffee/lib/abstract-chosen.coffee
+++ b/coffee/lib/abstract-chosen.coffee
@@ -88,6 +88,7 @@ class AbstractChosen
     option_el.style.cssText = option.style
     option_el.setAttribute("data-option-array-index", option.array_index)
     option_el.innerHTML = option.search_text
+    option_el.title = option.title if option.title
 
     this.outerHTML(option_el)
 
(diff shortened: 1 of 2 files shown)


optimatika:
Is it correct that this has not (yet) been accepted?

We're currently using a patched version of chosen, but would very much prefer to use an officially released version.

Is there a problem?

sdarisi:
Is this pull request is ever going to be reviewed and merged? We are using a patched version of Chosen as well, it'd be nice to see these changes in official version and use it instead.

metaist:
@harvesthq/chosen-developers can you please help these nice people? It's 4 lines of code.

optimatika:
Thanks!

PR #1705, opened 2013-12-28
Call show_search_field_default after a result is selected.

### LINKED ISSUE #1701: Remove default class after an option is chosen in multiple select
On multiple select initialized without selected values there is a `default` class on `search-field input[type=text].default`

After I choose a value for the first time the class `default` isn't being removed until click out of the control or use backspace to remove the chosen item.

It's a rather small issue but If you have the `default` class styled in a way to match the default look of your other controls, you will want this class to be removed immediately after an item is chosen.

pfiller:
@harvesthq/chosen-developers 

#1701 is kind of a non-issue in my eyes, but the change is simple enough and I actually prefer this to what exists now. After selecting a result, we should always be resetting a field's value the same way. This adds an extra class removal call in the non-multiple version, but it's harmless.

What say you?

Fixes #1701

### DIFF
--- a/coffee/chosen.jquery.coffee
+++ b/coffee/chosen.jquery.coffee
@@ -351,8 +351,7 @@ class Chosen extends AbstractChosen
         this.single_set_selected_text(item.text)
 
       this.results_hide() unless (evt.metaKey or evt.ctrlKey) and @is_multiple
-
-      @search_field.val ""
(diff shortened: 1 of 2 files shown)


pfiller:
I thought it was the right thing to do in December of 2013 and I think it's the right thing now. Merging!

PR #1731, opened 2014-01-17
Calling activate_field on choices_click, fixes #1729

### LINKED ISSUE #1729: Multi Select: `.search-field input` width does not fill width of `.chosen-choices`
For **Multiple Select**, the width of `.search-field input` does not fill the width of the _perceived_ input of `.chosen-choices`.

This is illustrated clearly in the following screenshot from the Chosen demo page at http://harvesthq.github.io/chosen/

screen shot 2014-01-15 at 6 47 43 pm

This really becomes a problem when a user attempts to click inside the _perceived_ input field after the field has been given focus. This actually causes the _real_ input field to lose focus and potentially causes a lot of confusion for the user when clicking inside what looks like an input field doesn't actually give it focus.

It looks like there was a small mention of this issue in this thread (https://github.com/harvesthq/chosen/pull/407#issuecomment-3915749), but it doesn't look like it was ever truly addressed.
#### Steps to reproduce:
1. Click inside chosen multi select form to open dropdown and give focus to the input field
2. Click the _perceived_ input field near the right edge.
3. Try to type, realize the input doesn't actually have focus.
#### Possible Solutions
1. At first glance, it looks like a true solution would modify the inputs width depending on the width of `.chosen-choices` minus the width of any already selected values. We probably want to give the input some minimum width also, to prevent it from becoming confusingly narrow.
2. Alternatively, I guess you could just focus on the child `input` on any click events on `.chosen-choices`.

Is there any interest in merging a solution to this issue? I'd be happy to cook up a pull request with either of the above solutions.

jordanstephens:
Fixes #1729 simply by sending focus to `.search-field input` when clicking `.chosen-choices`.

At first, I imagined two possible solutions to this issue:
1. Modify the width of `.search-field input` to fill the width of `.chosen-choices` minus the width of any currently selected values.
2. Send focus to `.search-field input` when clicking `.chosen-choices`.

The first option was not selected as it was more complicated and it would fail to address the case of clicking inside the red circled area in the following screenshot:
screen shot 2014-01-16 at 7 20 14 pm

### DIFF
--- a/coffee/lib/abstract-chosen.coffee
+++ b/coffee/lib/abstract-chosen.coffee
@@ -202,6 +202,7 @@ class AbstractChosen
 
   choices_click: (evt) ->
     evt.preventDefault()
+    this.activate_field()
     this.results_show() unless @results_showing or @is_disabled
 
   keyup_checker: (evt) ->


koenpunt:
@harvesthq/chosen-developers, @stof; got anything against merging this?

anteq:
+1, we're affected by #1729 and this fixes the problem - would love to see this merged! @harvesthq/chosen-developers

jordanstephens:
Wow, this was on of the first pull requests I ever submitted to an open source project—a little over three years ago. Thanks for merging this!

PR #1776, opened 2014-02-12
Allow submit with enter if results not showing

### LINKED ISSUE #434: Pressing enter doesn't work to submit?
Hey guys,

Great work on Chosen! We absolutely love it.

We just integrated it into one of our forms, which has _only_ a `<select>` dropdown and an `<input type="submit">` button. Unfortunately, we're seeing that pressing enter to submit the form no longer works, when the Chosen dropdown has focus.

We haven't tested this on other browsers yet, but we're running Chrome 16 on Mac.

Here's the simple HTML:

``` html
<form action="" method="POST">
    <select name="user[id]" required class="chosen"
            data-placeholder="Choose a person…"
            data-no-results-text="No one named">
        <option value=""></option>
        <option value="1">Daniel Gasienica</option>
        <option value="2">Aseem Kishore</option>
        <!-- ... -->
    </select>

    <input type="submit" class="button" value="Suggest">
</form>
```

Thanks in advance for your consideration!

danez:
If the result layer is not open an the user presses enter the from should be submitted.

Taken from #434, which seems to work already for a bunch of people.

Fixes #434

### DIFF
--- a/coffee/chosen.jquery.coffee
+++ b/coffee/chosen.jquery.coffee
@@ -463,7 +463,7 @@ class Chosen extends AbstractChosen
         @mouse_on_container = false
         break
       when 13
-        evt.preventDefault()
+        evt.preventDefault() if this.results_showing
         break
       when 38
         evt.preventDefault()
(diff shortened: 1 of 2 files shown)


hoshomoh:
documentation for this isn't updated on the website.

tjschuck:
@hoshomoh How would you like this documented?  This isn't really a usage thing...

PR #1806, opened 2014-03-03
escape html when setting selected item to prevent xss

meleyal:


pfiller:
#2254 does this in a way that works for both versions of Chosen. Going to close this in favor of that solution.

Thanks, @meleyal.

PR #1882, opened 2014-04-09
no_results_fallback_option_index

gazdanebojsa:
Gives possibility to have one `<option>` from the original `<select>` chosen as a fallback default. 

If the user filters out all other `<option>`s, then this `<option>` will remain selected and highlighted. On `blur`, it will remain selected. 

Particularly useful for cases where a value is required and "Other" is valid `<option>`.

chosen-no_options_fallback


pfiller:

Sorry this has sat open so long @gazdanebojsa. The hope is that by getting things under control, we can get to a place in the future where we can once again start adding new functionality (at which point we'd come back to closed PRs like this). Thanks for your patience and understanding!

PR #1902, opened 2014-04-22
Prevent search field losing focus

itimeheke:
First click to chosen container works fine (search_field gets focus), but when chosen is already active(this.results_showing == true) then input loses focus unless clicked on the search field itself


pfiller:
I'm not able to recreate the issue so I'm going to close this PR un-merged. If you'd like us to reconsider, please link to a publicly accessible page (or a Fiddle) that shows the issue so we can attempt to reproduce. Providing steps to make it happen helps also!

Thanks, @itimeheke!

Issue #1660, opened 2013-11-18
Feature: Adding an option for search by option.value


MasonThornberg:
I was creating a dropdown where users would be able to search by both the text and value of <option> items.
For example: <option value="customerId">Customer Name</option>, witht the ability to narrow results based on customerId and Customer Name

I was able to modify chosen.jquery.js for my own purposes, but I am unfamiliar with coffeescript and the whole git process, so I thought I would offer this up if anyone who is already making improvements would like to add this as well.

chosen.jquery.js:

```
AbstractChosen.prototype.set_default_values = function() {
    ... other code ...
    //Added optional parameter for optional search values
    this.search_option_values = this.options.search_option_values || false;
    ... other code ...
}

AbstractChosen.prototype.winnow_results = function() {

  ... other code ...

    if (!(option.group && !this.group_search)) {
       ... code that deals with regular match on <option>text</option> ... 
    }

    //Beginning of changes
    //Try searching option.value, if true, does the same stuff as above, just without the text underline stuff.
    else if (this.search_option_values && this.search_string_match(option.value, regex))
        option.search_match = true;
        results += 1;
        if (results_group != null) {
            results_group.group_match = true;
        }
    }
    //End of changes
    .. other code ...
}
$(".choosen-dropdown").chosen({ search_option_values: true });
```

PR #1923, opened 2014-05-05
Adding Issue #1660 to Coffee-script

stephandesouza:
* Feature: Adding an option for search by option.value


pfiller:

Sorry this has sat open so long @stephandesouza. The hope is that by getting things under control, we can get to a place in the future where we can once again start adding new functionality (at which point we'd come back to closed PRs like this). Thanks for your patience and understanding!

Issue #800, opened 2012-09-05
Search by text and value


akaroot:
Hi!
Please add capability to search in select lists not only by text but  by value too.

For example, 

```
  <select multiple class='chzn-select'>
                 <option value="some_value"> text one </ option>
                 <option value="other_value"> text 2 </ option>
                 ...
</ select>
```

And if I type "some" it should suggest me "text one".

9IPHP:
Is this feature done?

pfiller:
#628 is worth looking at, maybe. Would add value, but would be helpful.

ecrofeg:
Going on with this one #2796

PR #1925, opened 2014-05-05
Allow filter with data-* attributes

stephandesouza:
I created a new option "search_data_attributes" that lets you define which fields will be used for filtering based on #800


Nijikokun:
Can someone look at this? This would be amazing.

pfiller:

Sorry this has sat open so long @stephandesouza. The hope is that by getting things under control, we can get to a place in the future where we can once again start adding new functionality (at which point we'd come back to closed PRs like this). Thanks for your patience and understanding!

PR #1971, opened 2014-06-17
Fix for issues/1595

k-zelenkov:
It makes possible search with 'group_search:false' option and print group's label instead 'undefined'


pfiller:
This commit got behind master due to the fact that it's been open since June. Sorry about that.

The new PR #2255 contains the fix and credits @k-zelenkov. Thanks!

PR #1973, opened 2014-06-17
Allow native select interface for iOS (including iPad) and Android—potential improvement over #1388

jordanstephens:
Using chosen on small touch screens can be difficult and can feel clumsy. It looks like this was previously discussed in #1388, but the solution was to totally disable chosen for iPhones, iPods, and Android.

I think we can do a little better than that by combining the superior inactive-state UI of chosen with the superior active-state UI of the mobile device browser's select field interface. Not to mention the other features of chosen, [such as `max_selected_options`](https://github.com/harvesthq/chosen/pull/1388#issuecomment-37421975), that are lost right now by completely disabling chosen for mobile users.

We can accomplish this through a sort-of trick where a form element's opacity can be set to 0 so that it is unseen while still remaining interactive.

In this case, we can render the `.chosen-select` node in front of the `.chosen-choices` or `.chosen-single` node, but with an opacity of 0, so that it is invisible but interactive.

img_0003
img_0005

This *does* come at the expense of search on iPads, but I think the native experience so far outweights the default chosen experience here that it is an acceptable compromise.

Additionally, all of this behavior can be toggled with a `use_native_interface` option so the current (non-native) behavior could be forced if desired. However, I have configured it to be enabled by default for iOS (including iPads) and for Android through the same userAgent string matching that was previously used to disable chosen on those devices.

This does require one major change to the markup structure: `.chosen-select` is now nested inside of `.chosen-container` rather than being a previous sibling to it. I did this so I could use `.chosen-container` as a positioning anchor for `.chosen-select`.

Happy to make adjustments if you think anything could be better. Thanks!

### DIFF
--- a/coffee/chosen.jquery.coffee
+++ b/coffee/chosen.jquery.coffee
@@ -69,35 +69,36 @@ class Chosen extends AbstractChosen
     @form_field_jq.trigger("chosen:ready", {chosen: this})
 
   register_observers: ->
-    @container.bind 'touchstart.chosen', (evt) => this.container_mousedown(evt); return
-    @container.bind 'touchend.chosen', (evt) => this.container_mouseup(evt); return
+    unless this.use_native_interface()
+      @container.bind 'touchstart.chosen', (evt) => this.container_mousedown(evt); return
+      @container.bind 'touchend.chosen', (evt) => this.container_mouseup(evt); return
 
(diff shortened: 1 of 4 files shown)


pfiller:

Sorry this has sat open so long @jordanstephens. The hope is that by getting things under control, we can get to a place in the future where we can once again start adding new functionality (at which point we'd come back to closed PRs like this). Thanks for your patience and understanding!

PR #1990, opened 2014-06-27
Fixes #1410 - page scrolling to the left when navigating or using a single drop down

### LINKED ISSUE #1410: Using a chosen-element always fires a horizontal page scroll to the left

lukeapage:
Fixes the horizontal scrolling of the page when scrolled and mouse navigating or keyboard navigating to a drop down.

replacement fix instead of https://github.com/harvesthq/chosen/pull/1436

### DIFF
--- a/sass/chosen.scss
+++ b/sass/chosen.scss
@@ -17,7 +17,7 @@ $chosen-sprite-retina: image-url('chosen-sprite@2x.png') !default;
   .chosen-drop {
     position: absolute;
     top: 100%;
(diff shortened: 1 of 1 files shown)


pfiller:
Thanks @lukeapage, but this doesn't seem to be solving the problem for us. In Chrome, the dropdown's contents are visible even when the select isn't active. You're welcome to re-open if you are able to fix that.

PR #1994, opened 2014-06-27
Fixes focus highlighting on disabled chosen select boxes (Chrome, Windows)

oriondean:
Clicking on a disabled chosen element focuses it and shows a highlight border around the element. 
You can view an example here: http://jsfiddle.net/HpACF/

This is caused by tabindex -1 being added to the chosen-single anchor tag.
Any anchor tag without an href attribute cannot be tabbed to or focused into. 
By adding tabindex -1 to the anchor tag, it means that it can not be focused into via tabbing.

My fix removes this tabindex -1 attribute from the anchor tag. 
This means that a disabled chosen select box is no longer focusable when disabled, removing the highlight border when clicking into the chosen element.


pfiller:
Thanks, @oriondean. This looks good to me and I think I'd like to get it merged before this PR sits around for the full year.

PR #1997, opened 2014-06-30
=BG= stop event bubbling in Chosen.result_select. Fix #1996

### LINKED ISSUE #1996: Upon tapping a selection from the dropdown, the input element directly underneath gets focus
Affects iOS Safari.

When you have a chosen select, followed by a number of standard `<input>` fields directly below it:

- Tap the chosen select
- Dropdown appears, covering the standard `<input>` elements below it
- Tap one of the entries in the dropdown
- That entry gets selected, and the dropdown is hidden
- The `<input>` element directly underneath the dropdown entry that was tapped gets focus

The last step in this sequence in the problem. It should not happen. It seems that the event listener does not handle the event, and instead let's it bubble up the DOM hierarchy.

bguiz:
Stop event bubbling in Chosen.result_select. Fix https://github.com/harvesthq/chosen/issues/1996

### DIFF
--- a/coffee/chosen.jquery.coffee
+++ b/coffee/chosen.jquery.coffee
@@ -329,6 +329,8 @@ class Chosen extends AbstractChosen
     @selected_item.find("abbr").remove()
 
   result_select: (evt) ->
+    evt.stopPropagation()
+    evt.preventDefault()
     if @result_highlight
       high = @result_highlight
 


pfiller:
I'm pretty sure that #2093 made this unnecessary. Thanks, @bguiz.

PR #2007, opened 2014-07-07
Add right padding when allow_single_delect=true

AlJohri:
The "x" button covers up the text when allow_single_deselect=true. For example, if my options are "yes" and "no", the options are entirely covered by the "x".

Increasing the width by 25px fixes this problem for all cases.


pfiller:
This is the same as #2589, which I just [closed un-merged](https://github.com/harvesthq/chosen/pull/2589#issuecomment-223434048). My comment there applies just the same here.

Thanks anyways, @AlJohri, and sorry for the slowwwww review.

PR #2017, opened 2014-07-18
Fix IME on Firefox.

jiru:
Hello, this pull request implements the IME part of #1946, in a hopefully proper way.

> Firefox prevents keystrokes from being detected while inputting with
an IME. That is, keystrokes produced from the start and the end of an
IME composition and its ending are blocked. As a result, pressing the
enter key to commit a composition does not filter the results and
select the previously highlighted result right away, preventing Firefox
users from ever changing the seleted item.

> To work around this problem, we differenciate “pressing enter to commit
a composition” from “pressing enter to select the highlighted result in
Chosen” by checking if the keydown event preceded the keyup one.

> Chrome and IE are not affected by this problem because they [do not
block key events during composition](https://bugzilla.mozilla.org/show_bug.cgi?id=354358#c40). If Firefox happen to switch to
Chrome and IE’s beheaviour, this commit should be reverted.

### DIFF
--- a/coffee/chosen.jquery.coffee
+++ b/coffee/chosen.jquery.coffee
@@ -466,6 +466,7 @@ class Chosen extends AbstractChosen
         break
       when 13
         evt.preventDefault() if this.results_showing
+        @enter_was_keyed_down = true
         break
       when 38
         evt.preventDefault()
--- a/coffee/chosen.proto.coffee
+++ b/coffee/chosen.proto.coffee
@@ -466,6 +466,7 @@ class @Chosen extends AbstractChosen
         break
       when 13
         evt.preventDefault() if this.results_showing
+        @enter_was_keyed_down = true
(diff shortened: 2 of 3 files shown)


pfiller:
This is super old and I think I'm going to close it un-merged. It seems like someone from Mozilla [acknowledged](https://bugzilla.mozilla.org/show_bug.cgi?id=354358#c43) that they will handle `keydown` and `keyup` events the way Chrome and Safari handle them:

> We'll dispatch "keydown" and "keyup" events in the future for compatibility with other browsers

This fix would still work for the meantime, but I'm hesitant to add a relatively gnarly workaround for something so few people have run into (the four year old Chosen issue has 3 commenters). Hopefully anyone hitting it now will find this thread and work the fix in.

Thanks anyways, @jiru.

PR #2030, opened 2014-08-04
Restore user input

Piccirello:
These changes save any user input in a new variable named search_input. The user input is restored to the search field whenever the field regains focus, including after a result has already been selected. This allows the user to more easily correct their selection or continue where they left off should the field lose focus for any reason (either accidental or purposeful).

If this is not the desired default behavior, I do believe that this would be beneficial to have as an option.

### DIFF
--- a/coffee/chosen.jquery.coffee
+++ b/coffee/chosen.jquery.coffee
@@ -234,7 +234,7 @@ class Chosen extends AbstractChosen
     @results_showing = true
 
     @search_field.focus()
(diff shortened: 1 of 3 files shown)


pfiller:
Thanks @Piccirello 

I don't think this behavior is what a person would expect to happen after choosing a result. Unless you were adding a lot of similarly named things, it would be a little frustrating.

An interesting idea, though. Thanks for sharing.

PR #2058, opened 2014-08-26
Custom Container Tags for Inter-plugin Compatability

SamJBarney:
Using the plugins Chosen and NestedSortable together produces a
collision and breaks NestedSortable.  This collision is because of the
rendered tags of Chosen hooked by NestedSortable.  Given the choice
between patching jQuery UI and Chosen, Chosen was patched to use a set
of customizable tags.  The following options are now available:

- outer_container -- default value = 'ul'

- inner_container -- default value = 'li'

The stylesheet has no references to the HTML tags (e.g. li.search-
choice is now .search-choice) including the stand-alone tags through
clever use of wildcards.  The styling will apply to all tags used.
This is naturally dangerous, as a set of inappropriate tags could be
forced.  However, the ability to change 'ul' and 'li' to 'div' and
'span' respectively, thus resolving the conflict between libraries
without affecting either's functionality far outweighs the fact that
an individual can break his or her markup.

### DIFF
--- a/coffee/chosen.jquery.coffee
+++ b/coffee/chosen.jquery.coffee
@@ -40,22 +40,22 @@ class Chosen extends AbstractChosen
     @container = ($ "<div />", container_props)
 
     if @is_multiple
-      @container.html '<ul class="chosen-choices"><li class="search-field"><input type="text" value="' + @default_text + '" class="default" autocomplete="off" style="width:25px;" /></li></ul><div class="chosen-drop"><ul class="chosen-results"></ul></div>'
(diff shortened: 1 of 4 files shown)


pfiller:
I agree with @stof that this isn't a great fit for Chosen. I don't doubt that some people will find a reason to swap out their own tags, but we try to avoid edge case behaviors becoming fully supported features in Chosen.

Every feature comes with the burden of maintenance. I can imagine a scenario where we make a change to styles with unordered lists in mind and it breaks someone else who decided that a definition list was better. Now we have to figure out how to update our styles to support both – and any other HTML someone chooses.

I appreciate your effort @SamJBarney and I hope that someone looking for this functionality will find this PR. At the very least, we'll be able to direct them this way.

Thanks!

PR #2076, opened 2014-09-10
allow hiding selected option for single selects

tmont:
I needed to hide selected options in the dropdown for single selects. I'm not sure why it was explicitly stated that it only works for multiple selects, because it seems like a useful option for single selects as well.

This is important for my use case because after reskinning the Chose UI, showing the selected option in the dropdown was distracting:

Before:
screen shot 2014-09-10 at 10 18 47 am

After:
screen shot 2014-09-10 at 10 19 01 am

The selected option being in the dropdown is a little distracting from a UI perspective, so it's convenient to be able to hide it.

This might warrant a new option (`display_selected_option_single` or something), but it shouldn't affect any current code, since it was ignored on single selects anywa.

### DIFF
--- a/coffee/lib/abstract-chosen.coffee
+++ b/coffee/lib/abstract-chosen.coffee
@@ -232,7 +232,7 @@ class AbstractChosen
     return if @options.width? then @options.width else "#{@form_field.offsetWidth}px"
 
   include_option_in_results: (option) ->
-    return false if @is_multiple and (not @display_selected_options and option.selected)
+    return false if not @display_selected_options and option.selected
     return false if not @display_disabled_options and option.disabled
     return false if option.empty
 


pfiller:
This is a select box that has the selected option hidden:

confusing select

If you were a regular person (that is, not a power-user-dev-ninja), where would you click to make sure that the United States _stays_ selected? No matter what answer you give: you'd be guessing.

Here's normal chosen:

not confusing

Answering the same question is pretty easy, huh?

Anywho ... that's why it's not included for single and why I don't think I'll change it today. I can totally understand if you disagree with my reasoning and that's why Chosen is MIT licensed!

Thanks, @tmont. I do enjoy seeing screenshots of Chosen in "night mode"!

PR #2091, opened 2014-09-27
Remove forced widths in order to get responsive dropdown menus

elboletaire:
As the title suggests, I've removed the forced widths used when creating chosen objects. Now the width of the select menus depends on their wrapper.

I've uploaded it to `gh-pages` so you can see it working: http://elboletaire.github.io/chosen/

Obviously, resize the window to see it in action :smiley:

### DIFF
--- a/coffee/chosen.jquery.coffee
+++ b/coffee/chosen.jquery.coffee
@@ -32,7 +32,6 @@ class Chosen extends AbstractChosen
 
     container_props =
       'class': container_classes.join ' '
-      'style': "width: #{this.container_width()};"
       'title': @form_field.title
 
     container_props.id = @form_field.id.replace(/[^\w]/g, '_') + "_chosen" if @form_field.id.length
--- a/coffee/chosen.proto.coffee
+++ b/coffee/chosen.proto.coffee
@@ -20,7 +20,6 @@ class @Chosen extends AbstractChosen
 
     container_props =
       'class': container_classes.join ' '
-      'style': "width: #{this.container_width()};"
       'title': @form_field.title
 
     container_props.id = @form_field.id.replace(/[^\w]/g, '_') + "_chosen" if @form_field.id.length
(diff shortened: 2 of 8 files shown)


cfbauer:
Would love to see this merged, it allows us to use our own width where necessary.

elboletaire:
I would love it too... :sweat_smile:

joerideg:
Would love to see this merged. Style should not be determined via the js or html. At least set no width style by default if width is not provided as an option.

pfiller:
Chosen specifically tries to match the width of the select it's replacing. The reason for this is to match the behavior browsers exhibit when sizing selects (which is to make the select the appropriate width for the content). In most cases, we've preferred to keep Chosen's behavior in line with what the browser will do and width is not going to be an exception.

Browsers offer you a way to work around the default behavior by using css ... and so does Chosen! You can also specify a width for Chosen to use. If you'd like Chosen to be 100%, you just need to specify a width of 100%:

```JavaScript
$("select").chosen({ width: 100% });
```

Thanks, @elboletaire

pfiller:
The above example is missing quotes:

```JavaScript
$("select").chosen({ width: '100%' });
```

elboletaire:
Well, that's not the answer I was looking for ... and I'm sure it's neither the solution for the people who commented this PR.

I'll do my own fork then. Thanks for your help :)

cfbauer:
@pfiller It's my understanding that Chosen sets widths inline. Is that correct? If so, setting the width with css would require the !important hack which isn't a good practice.

cfbauer:
To add: It's rare I would want to universally set the width of select boxes to 100%. Most of the time I would use media queries in css to make select boxes 100%, which would get tricky fast having to use !important.

joerideg:
So @pfiller if chosen is trying to match the width of the select element, and width is not specitied in the options object, does that mean I can set width styles on the select with css and chosen will (on pageload) use that width?

pfiller:
> It's my understanding that Chosen sets widths inline. Is that correct? If so, setting the width with css would require the !important hack which isn't a good practice.

You can set the width of the select in CSS and Chosen will try to match!

> if chosen is trying to match the width of the select element, and width is not specitied in the options object, does that mean I can set width styles on the select with css and chosen will (on pageload) use that width?

Yes, as long as the width is in px. If you need a percentage width, you need to specify it in the options object.

PR #2093, opened 2014-09-28
fix that touching an item in the result list will also touch the element underneath it

### LINKED ISSUE #1816: Chosen has issues when option is selected on iPad
When you select an option on iPad the element behind the item list also receives the event (so let's say there is a checkbox underneath the item I'm selecting, the checkbox will be checked, which is not desirable)

### LINKED ISSUE #1887: iPad : Select option triggers link below
Hi, I'm getting the same issue as described below (on <a href="http://stackoverflow.com/questions/22520164/jquery-chosen-select-option-triggers-link-with-lower-z-index-when-selected-on-ip">stackoverflow 19 march</a>).
Tapping on an option selects it, but the tap event seems to continue, and after the select list is hidden, it triggers a link that was below the selected option.

<b>Stackoverflow issue description</b>
"I am using jQuery Chosen plugin to style select box #foo. I have a block level link (#bar) which is positioned just below #foo and which has a lower (0) z index. #foo 's options have a z-index of 1010.

On my ipad mini, when I select an option from #foo, the link (#bar) is triggered. I have no issues on an android tablet, and I am not having any problems on any non-touch browser. Just iPad

Here is my code:

```html
<select id="foo">
    <option>foo</option>
    <option>foo</option>
    <option>foo</option>
    <option>foo</option>
    <option>foo</option>
    <option>foo</option>
    <option>foo</option>
    <option>foo</option>
    <option>foo</option>
    <option>foo</option>
    <option>foo</option>
    <option>foo</option>
    <option>foo</option>
</select>

<a href="http:www.stackoverflow.com" id="bar">GO!</a>
```
and the CSS

```css
#foo {width:300px;}

#bar {display:block; width:400px; height:100px; color:#FFF; text-align:center; font-size:30px; background:#FF0000; line-height:90px; z-index:0;}
```

<a href="http://jsfiddle.net/356xX/4/">here</a> is my fiddle. The css for jQuery chosen is in the fiddle's external resources.
Thanks for help!

craue:
This fixes #1816 (and fixes #1887).

Props to @jonsprague0000 for [the fix](https://github.com/harvesthq/chosen/issues/1816#issuecomment-37178481). I'm just creating the PR to finally push it into the codebase.

### DIFF
--- a/coffee/chosen.jquery.coffee
+++ b/coffee/chosen.jquery.coffee
@@ -360,6 +360,9 @@ class Chosen extends AbstractChosen
 
       @form_field_jq.trigger "change", {'selected': @form_field.options[item.options_index].value} if @is_multiple || @form_field.selectedIndex != @current_selectedIndex
       @current_selectedIndex = @form_field.selectedIndex
+
+      evt.preventDefault()
+
       this.search_field_scale()
 
   single_set_selected_text: (text=@default_text) ->
--- a/coffee/chosen.proto.coffee
+++ b/coffee/chosen.proto.coffee
@@ -356,6 +356,8 @@ class @Chosen extends AbstractChosen
       @form_field.simulate("change") if typeof Event.simulate is 'function' && (@is_multiple || @form_field.selectedIndex != @current_selectedIndex)
       @current_selectedIndex = @form_field.selectedIndex
 
+      evt.preventDefault()
+
       this.search_field_scale()
 
   single_set_selected_text: (text=@default_text) ->


pauldcollins:
Hi guys, thanks for fixing this. 

Will this automatically go into the latest release? I've just grabbed version 1.3.0 and tried, but it doesn't seem to have solved the problem for me.

tjschuck:
@pauldcollins This will be included with the next release (which as you noticed, has not yet been released).

In the meantime, you can build Chosen yourself from source by [following these instructions](https://github.com/harvesthq/chosen/blob/master/contributing.md#using-coffeescript-and-grunt).

andreasbeer1981:
:+1: great fix!

Issue #2004, opened 2014-07-04
When standard placeholder text and selecting several times, last selection not possible to select


utkant:
Hi!

I am using chosen and have a selectbox with two options. When chosen is shown for the first time it shows the two options and the standard 'choose an option' text. If I then select first the first choice, then the second choise I get an error message saying: 

"TypeError: this.form_field.options[item.options_index] is undefined" in line 1002 in chosen.jquery.js.

The version in that file is 1.1.0.

Is this a known issue?

pfiller:
I'm not able to recreate this issue, so I'm going to close it. It's quite old so it's possible it was fixed in an unrelated PR.

If you are still experiencing the issue, please link to a publicly accessible page (or a [Fiddle](http://jsfiddle.net)) that shows the issue so we can attempt to reproduce. Providing steps to make it happen helps also!

PR #2097, opened 2014-10-02
fix: last selection not possible to select after first selection

rkmax:
fix the issue #2004


pfiller:
I wasn't able to recreate the issue (#2004) that this PR addresses. It's possible something else fixed it by accident. It's also possible it's still a problem -- if so, please provide a link to a publicly accessible page (or a [Fiddle](http://jsfiddle.net)) that shows the issue so we can attempt to reproduce. Providing steps to make it happen helps also!

Thanks @rkmax!

PR #2101, opened 2014-10-06
Always open mode for Chosen

liorch88:
I added an "always open" mode. By specifying "always_open: true" in the Chosen construction, it will make Chosen always open - immediately, on instansiation, and will remain open after selecting an option.


cris:
:+1:  - need this feature for current task

pfiller:
I don't think this kind of functionality should be baked into Chosen and I'd go further to say that it's possible to implement something like this using existing options in Chosen. After you've instantiated Chosen (or an option has been selected), you can trigger `chosen:open.chosen` to achieve nearly the same behavior.

Thanks anyways, @liorch88

PR #2119, opened 2014-10-22
Fix touch event propagation

DaanDD:
Fix a bug where the chosen element immediately closes again after opening it with a touch event, because both the touch and mouse events get triggered the first time.


bpaul:
:+1:

bpaul:
Would love to see this merged and released. Fixes a huge problem on mobile devices...

pfiller:
Thanks @DaanDD. I've confirmed that this fixes #2171.

Issue #73, opened 2011-07-26
XSS injections vulnerability


lostintime:
the plugin does not filter text from source selects: for option with text like: "abc &quot;&gt;&lt;script&gt;alert(&#039;hack&#039;)&lt;/script&gt; def &quot;&gt;&lt;script&gt;alert(&#039;hack!&#039;)&lt;/script&gt; "
it will execute the script inside (on init and options change).

PR #2157, opened 2014-12-05
Allows searching for & and other special characters

lode:
This allows me to search for special characters like ampersand in select boxes. I have a client which has trouble without this as they have large selects and a few options are only findable via ampersands. (i.e. `a&b` will not be found easily as `a` and `b` are far too common).

I don't know if this might break something (why the `html()` was needed), but for me it seems to work.

### DIFF
--- a/coffee/chosen.jquery.coffee
+++ b/coffee/chosen.jquery.coffee
@@ -398,7 +398,7 @@ class Chosen extends AbstractChosen
     @selected_item.addClass("chosen-single-with-deselect")
 
   get_search_text: ->
-    if @search_field.val() is @default_text then "" else $('<div/>').text($.trim(@search_field.val())).html()
+    if @search_field.val() is @default_text then "" else $('<div/>').text($.trim(@search_field.val())).text()
 
   winnow_results_set_highlight: ->
     selected_results = if not @is_multiple then @search_results.find(".result-selected.active-result") else []
--- a/coffee/chosen.proto.coffee
+++ b/coffee/chosen.proto.coffee
@@ -393,7 +393,7 @@ class @Chosen extends AbstractChosen
     @selected_item.addClassName("chosen-single-with-deselect")
 
   get_search_text: ->
-    if @search_field.value is @default_text then "" else @search_field.value.strip().escapeHTML()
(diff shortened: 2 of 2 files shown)


rkhmelichek:
I have the same issue, but we're using the prototype version of Chosen.
I opened up a pull request for it here: https://github.com/harvesthq/chosen/pull/2186

tjschuck:
> I don't know if this might break something (why the `html()` was needed)

This function was added in https://github.com/harvesthq/chosen/pull/1339 by @pfiller

tjschuck:
@lode Can you cherry-pick @rkhmelichek's commit over to your PR so this is just one PR?

lode:
@tjschuck, I'll do.

By the way, I don't think it was added by #1339, it was just moved, see https://github.com/harvesthq/chosen/pull/1339/files#diff-c82fb80b20c6f4dcdec5a08a57059c30L409. As far as I know it was added in https://github.com/harvesthq/chosen/commit/aed8ba82bd889b9b55f25cc6d06f43142f3eeb1c.

tjschuck:
@lode Good call -- that was in https://github.com/harvesthq/chosen/pull/89, which refs https://github.com/harvesthq/chosen/issues/73

lode:
@tjschuck, how do you prefer the merges to be done?
Shall I merge in master from harvesthq? Or do you rather have a `rebase`?

By the way, if you feel that this creates xss trouble again, please let me know.

tjschuck:
Rebasing should be fine.

I can't give the +1 on the code change -- I'm mostly just a babysitter for Chosen, managing/triaging issues.  One of the @harvesthq/chosen-developers will have to assess whether this will cause XSS or other issues.

lode:
I see @tjschuck. I added @rkhmelichek's commit (thanks @rkhmelichek!).

pfiller:
I don't think this is an issue as of 1.4.0 as evidenced in this [fiddle](http://jsfiddle.net/ne4gkkdp/). Perhaps the merging of #2254 made it all better?

In any event, I'm going to close this in the meantime. Thanks, @lode!

PR #2162, opened 2014-12-08
Add the original option's value to the created option's EL as data-value

liorch88:
To easily know what the original option value is for each option, without having to check the real, hidden select element (this was needed for my use case) I added the original option's value to the option element under "data-value".


pfiller:
@liorch88 I'm not 100% sure I see a need for this. Can you explain a little bit more about how you used it and why it might be useful to others? Thanks!

(closing in the meantime)

PR #2183, opened 2014-12-31
Add support for multiline search

gnicod:
When pasting several lines in the search field, if one or any line match
an option of the selectbox , those options will be selected.
The input:text is replaced by a textarea if this option is enabled


pfiller:
Thanks for sharing, @gnicod.

I don't think a solution like this is appropriate for Chosen at this time. It's a lot of code for something that seems like an uncommon behavior. I don't think a person would reasonably expect to paste in multiple lines and have all of those options auto-select. As much as we can, we try to keep Chosen's feature set closer to features that would be much more common.

PR #2194, opened 2015-01-14
Fixing placeholder bug

ComeBurguburu:
When someone fill the value of the placeholder the search stop working and show all result instead of "nothing found"

This pull request solved this issue

( I've closed my other pull request , this is the last one with only one commit /cc @tjschuck )

Issue #999, opened 2013-01-22
iPad Chrome - selecting an option - visual problem


amcsi:
When viewing any chosen select in Chrome on an iPad, when I select an option, instead of the expected highlighting the selected list element, and closing of the dropdown soon after, the selected list element does not get highlighted at all, and apparently the parent ul turns grey (showing that some tap event was registered on that element).

PR #2195, opened 2015-01-14
Fixed placeholder bug

ComeBurguburu:
when we set the value of placeholder in field Chosen returns all results

### DIFF
--- a/coffee/chosen.jquery.coffee
+++ b/coffee/chosen.jquery.coffee
@@ -398,7 +398,7 @@ class Chosen extends AbstractChosen
     @selected_item.addClass("chosen-single-with-deselect")
 
   get_search_text: ->
-    if @search_field.val() is @default_text then "" else $('<div/>').text($.trim(@search_field.val())).html()
+    $('<div/>').text($.trim(@search_field.val())).html()
 
   winnow_results_set_highlight: ->
     selected_results = if not @is_multiple then @search_results.find(".result-selected.active-result") else []
--- a/coffee/chosen.proto.coffee
+++ b/coffee/chosen.proto.coffee
@@ -393,7 +393,7 @@ class @Chosen extends AbstractChosen
     @selected_item.addClassName("chosen-single-with-deselect")
 
   get_search_text: ->
-    if @search_field.value is @default_text then "" else @search_field.value.strip().escapeHTML()
+    @search_field.value.strip().escapeHTML()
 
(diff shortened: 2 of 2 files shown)


tjschuck:
@stof @koenpunt #2194 has popped back up over here now :grinning:

koenpunt:
I'm good with this. @stof you still have any objections?

stof:
I think this would also require using a real placeholder for the field instead of setting the default text in the value

tjschuck:
@stof Is that a blocker for this PR, or just an additional enhancement to be made separately?

koenpunt:
I'll run it through multiple browsers, until now I've only tested in Chrome Canary

stof:
Well, I think the current implementation might use the placeholder text to run the search now in some cases

sergioviniciuss:
@stof  I think it was implemented by getting the 'value' instead of the placeholder tag to keep the support to older browsers like IE8

ComeBurguburu:
@sergioviniciuss is right !!!
the (fake)placeholder is set only when there are no selected item and the input is not focus, I guess that we can keep it as fake placeholder to keep compatibility.

@stof the value of the data-placeholder is only use once to change the color of the input (color: #999 like default HTML5 placeholder color) 
I didn't remove this condition.

but to run the search the placeholder is useless.
Placeholder is here to inform the user not the program.

ComeBurguburu:
This bug remain in 1.4.0 @pfiller 
http://jsfiddle.net/97huf6tn/

pfiller:
So, I'm _pretty_ sure that filtering out the default value is a relic of a time when we called `winnow_results` when Chosen was being set up. In other words, we were running the first "search" while the default value was sitting in the form field and we needed to match all elements. Was that stupid? Yes!  That's why we made that better at some point.

I don't think there's currently a way to have Chosen open and have the default text still living in the form field. If that's true, by the time `get_search_text` is called, we can assume the person using Chosen actually meant to type the value.

This is all a very long way of saying that I am :+1: on this change. Thanks, @ComeBurguburu!

ComeBurguburu:
Thanks  @pfiller 

May we remove this branch create by @koenpunt  because is merged now https://github.com/harvesthq/chosen/tree/ComeBurguburu-master ?

tjschuck:
@ComeBurguburu Branch deleted -- thanks!

PR #2200, opened 2015-01-19
Added a placeholder text for single select search box

liorch88:
I added a placeholder option for the search box (defaults to empty) so users could add a text such as "search now"


tjschuck:
Chosen already supports a default text placeholder -- see [the docs](http://harvesthq.github.io/chosen/#default-text-support).

liorch88:
It supports a placeholder for the select, not for the search box.
For example, in the docs you linked to, when you click the select it opens up with a blank search field. My addition adds a place holder, so you could write something like "type wanted value or a sub string of it"

tjschuck:
Aha.  Will reopen and mark as a feature request.  You'll have to await the @harvesthq/chosen-developers to determine whether it will be added.

Jessekevon:
Any updates on this feature? Also looking for a good solution for this.

pfiller:

Sorry this has sat open so long @liorch88. The hope is that by getting things under control, we can get to a place in the future where we can once again start adding new functionality (at which point we'd come back to closed PRs like this). Thanks for your patience and understanding!

Issue #2132, opened 2014-11-07
XSS vulnerability (new in v1.2.0)


philfreo:
Version 1.2.0 introduces a major XSS vulnerability

Version 1.1.0: no problem - http://jsfiddle.net/hxao6bc5/
Version 1.2.0: problem - http://jsfiddle.net/ptsL5md0/

![screenshot 2014-11-07 11 47 40](https://cloud.githubusercontent.com/assets/97612/4959322/57db4b00-66b7-11e4-965d-4a5db535db49.png)

Not sure of which commit introduces it, but here's [the diff](https://github.com/harvesthq/chosen/compare/v1.1.0...v1.2.0)

ograycode:
This is still an issue in v1.3.0, http://jsfiddle.net/wkm51zzm/

tjschuck:
Looks like this came in with https://github.com/harvesthq/chosen/pull/1638, which was held up for a while pending XSS issues discussed in https://github.com/harvesthq/chosen/issues/1150.

The conclusions of that may have been insufficient.  @stof was seemingly the sign-off on both of those, so maybe he has more to say here.  /cc @pfiller

Intrepidd:
This is a pretty important security issue, is someone working actively on this ?

PR #2209, opened 2015-01-26
Fix XSS on option content #2132

Intrepidd:
Patch suggestion by @obecker


stof:
This should be escaped only when inserting the text in the DOM, not when creating the ``text`` property here. Otherwise the ``text`` property does not contain text anymore but HTML.

pfiller:
I agree with @stof that we shouldn't be changing the text this way. We shoudln't be jamming the text into the dom without escaping, though. #2254 seems to handle that for us.

Thanks, @Intrepidd

PR #2217, opened 2015-02-02
Fix opt group label not appearing when triggering "chosen:updated" event

andreialecu:
If you changed the selection from code and triggered chosen:updated, the label for the optgroup wouldn't appear with `include_group_label_in_selected`. It only appears when clicked on manually by an user.

### DIFF
--- a/coffee/lib/abstract-chosen.coffee
+++ b/coffee/lib/abstract-chosen.coffee
@@ -77,7 +77,7 @@ class AbstractChosen
         if data.selected and @is_multiple
           this.choice_build data
         else if data.selected and not @is_multiple
-          this.single_set_selected_text(data.text)
+          this.single_set_selected_text(this.choice_label(data))
 
     content
 


andreialecu:
Note that the same problem may exist when working with a multiple selection box, I haven't checked/fixed that one, but it seems to be handled properly by the `choice_build` method.

tjschuck:
@koenpunt @pfiller This fixes an alleged behavioral bug introduced in https://github.com/harvesthq/chosen/pull/1610

andreialecu:
Since #1610 is not yet released, is there any other way I could create a jsfiddle to show this problem?

tjschuck:
You'd probably have to host it somewhere yourself to reference in the meantime.  You could even just put the compiled sources up on Gist and reference the raw version.

I'd rather not release the current version considering it has this issue in it -- fixing it first would be nice :)

andreialecu:
I have an easy repro here:
http://jsfiddle.net/5gm15hrg/

It doesn't have the new CSS so ignore that. Note that it shows Option 3 instead of Group 2: Option 3. If you click the option again it will show the correct text.

This PR fixes it.

pfiller:
Thanks, @andreialecu. This fixes the problem for me!

PR #2227, opened 2015-02-06
Add  tag name's check

ComeBurguburu:
- Check if selector is `<select>`
- Check if selector has at least one `<option>` child

PR #2231, opened 2015-02-10
Add tagname checker

ComeBurguburu:
Using class name as selector can provoque bug if the selector is not a `select` or if the select hasn't `option` child.

### DIFF
--- a/coffee/lib/abstract-chosen.coffee
+++ b/coffee/lib/abstract-chosen.coffee
@@ -295,4 +295,3 @@ class AbstractChosen
   @default_multiple_text: "Select Some Options"
   @default_single_text: "Select an Option"
   @default_no_result_text: "No results match"
-  
--- a/coffee/lib/abstract-chosen.coffee
+++ b/coffee/lib/abstract-chosen.coffee
@@ -284,11 +284,13 @@ class AbstractChosen
     
   @tagname_is_valid: : (@form_field) ->
   if @form_field.tagName.toUpperCase() != 'SELECT'
-    console.error 'Chosen error: invalid tagName ' + @form_field + ' is not a select'
+    if window.console
+      console.error 'Chosen error: invalid tagName ' + @form_field.selector + ' is not a select'
     return false
 
-  if @form_field.getElementsByTagName('option').length == 0
-    console.error 'Chosen error: ' + @form_field + ' has no options\'s childs '
+  if @form_field.options.length == 0
+      if  window.console
+       console.error 'Chosen error: ' + @form_field.selector + ' has no options\'s childs '
     return false
   return true
 
--- a/coffee/lib/abstract-chosen.coffee
+++ b/coffee/lib/abstract-chosen.coffee
@@ -281,16 +281,15 @@ class AbstractChosen
     if /Android/i.test(window.navigator.userAgent)
       return false if /Mobile/i.test(window.navigator.userAgent)
     return true
-    
-  @tagname_is_valid: (@form_field) ->
-  if @form_field.tagName.toUpperCase() != 'SELECT'
-    console?.error 'Chosen error: invalid tagName ' + @form_field.selector + ' is not a select'
-    return false
 
-  if @form_field.options.length == 0
-    console?.error 'Chosen error: ' + @form_field.selector + ' has no options\'s childs '
-    return false
-  return true
+  @tagname_is_valid: (@form_field) ->
+    if @form_field.tagName.toUpperCase() != 'SELECT'
+      console?.error 'Chosen error: invalid tagName ' + @form_field.selector + ' is not a select' 
+      return false
+    if @form_field.options.length == 0
+      console?.error 'Chosen error: ' + @form_field.selector + ' has no options\'s childs '
+      return false
+    return true
 
   @default_multiple_text: "Select Some Options"
   @default_single_text: "Select an Option"
--- a/coffee/lib/abstract-chosen.coffee
+++ b/coffee/lib/abstract-chosen.coffee
@@ -284,13 +284,11 @@ class AbstractChosen
     
   @tagname_is_valid: : (@form_field) ->
   if @form_field.tagName.toUpperCase() != 'SELECT'
-    if window.console
-      console.error 'Chosen error: invalid tagName ' + @form_field.selector + ' is not a select'
+    console?.error 'Chosen error: invalid tagName ' + @form_field.selector + ' is not a select'
     return false
 
   if @form_field.options.length == 0
-      if  window.console
-       console.error 'Chosen error: ' + @form_field.selector + ' has no options\'s childs '
+    console?.error 'Chosen error: ' + @form_field.selector + ' has no options\'s childs '
     return false
   return true
 
--- a/coffee/lib/abstract-chosen.coffee
+++ b/coffee/lib/abstract-chosen.coffee
@@ -282,7 +282,7 @@ class AbstractChosen
       return false if /Mobile/i.test(window.navigator.userAgent)
     return true
     
-  @tagname_is_valid: : (@form_field) ->
+  @tagname_is_valid: (@form_field) ->
   if @form_field.tagName.toUpperCase() != 'SELECT'
     console?.error 'Chosen error: invalid tagName ' + @form_field.selector + ' is not a select'
     return false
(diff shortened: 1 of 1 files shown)


ComeBurguburu:
clean pull request of #2227

coffee/lib/abstract-chosen.coffee
```diff
@@ -280,8 +281,19 @@ class AbstractChosen
     if /Android/i.test(window.navigator.userAgent)
       return false if /Mobile/i.test(window.navigator.userAgent)
     return true
+    
+  @tagname_is_valid: : (@form_field) ->
+  if @form_field.tagName.toUpperCase() != 'SELECT'
+    console.error 'Chosen error: invalid tagName ' + @form_field + ' is not a select'
```
stof: putting a DOM element in the message looks weird. How does the string representation looks like ?

coffee/lib/abstract-chosen.coffee
```diff
@@ -280,8 +281,19 @@ class AbstractChosen
...
+    return false
+
+  if @form_field.getElementsByTagName('option').length == 0
+    console.error 'Chosen error: ' + @form_field + ' has no options\'s childs '
```
stof: what if ``console.error`` does not exist (IE for instance) ?

coffee/lib/abstract-chosen.coffee
```diff
@@ -280,8 +281,19 @@ class AbstractChosen
     if /Android/i.test(window.navigator.userAgent)
       return false if /Mobile/i.test(window.navigator.userAgent)
     return true
+    
+  @tagname_is_valid: : (@form_field) ->
+  if @form_field.tagName.toUpperCase() != 'SELECT'
+    console.error 'Chosen error: invalid tagName ' + @form_field + ' is not a select'
+    return false
+
+  if @form_field.getElementsByTagName('option').length == 0
```
stof: use ``@form_field.options`` instead to get the options of a select

coffee/lib/abstract-chosen.coffee
ComeBurguburu: JQuery version  should be $(form_field).selector
don't know how to wrote it in coffee script

coffee/lib/abstract-chosen.coffee
ComeBurguburu: if (window.console) {

}

tjschuck: You can use [CoffeeScript's existential operator](http://coffeescript.org/#operators-existential) here instead:

```coffeescript
console?.error "..."
```

coffee/lib/abstract-chosen.coffee
```diff
@@ -280,8 +281,21 @@ class AbstractChosen
...
+    
+  @tagname_is_valid: : (@form_field) ->
+  if @form_field.tagName.toUpperCase() != 'SELECT'
+    if window.console
+      console.error 'Chosen error: invalid tagName ' + @form_field.selector + ' is not a select'
+    return false
+
+  if @form_field.options.length == 0
+      if  window.console
+       console.error 'Chosen error: ' + @form_field.selector + ' has no options\'s childs '
```
tjschuck: Grammar/formatting:  `' has no <option> children'`

coffee/lib/abstract-chosen.coffee
```diff
@@ -280,8 +281,19 @@ class AbstractChosen
...
+
+  if @form_field.options.length == 0
+    console?.error 'Chosen error: ' + @form_field.selector + ' has no options\'s childs '
+    return false
+  return true
 
   @default_multiple_text: "Select Some Options"
   @default_single_text: "Select an Option"
   @default_no_result_text: "No results match"
+
```
tjschuck: Please cut this extra line add.

coffee/lib/abstract-chosen.coffee
```diff
@@ -280,8 +281,18 @@ class AbstractChosen
...
+  if @form_field.options.length == 0
+    console?.error 'Chosen error: ' + @form_field.selector + ' has no options\'s childs '
+    return false
+  return true
 
   @default_multiple_text: "Select Some Options"
   @default_single_text: "Select an Option"
   @default_no_result_text: "No results match"
-
+
```
tjschuck: Heh, and now snip the extra indent addition :)

ComeBurguburu:
extra lines removed

coffee/lib/abstract-chosen.coffee
```diff
@@ -280,8 +281,17 @@ class AbstractChosen
...
       return false if /Mobile/i.test(window.navigator.userAgent)
     return true
+    
+  @tagname_is_valid: : (@form_field) ->
+  if @form_field.tagName.toUpperCase() != 'SELECT'
+    console?.error 'Chosen error: invalid tagName ' + @form_field.selector + ' is not a select'
+    return false
+
+  if @form_field.options.length == 0
+    console?.error 'Chosen error: ' + @form_field.selector + ' has no options\'s childs '
```
tjschuck: Grammar/formatting:  `' has no <option> children'`

coffee/lib/abstract-chosen.coffee
```diff
@@ -280,8 +281,17 @@ class AbstractChosen
     if /Android/i.test(window.navigator.userAgent)
       return false if /Mobile/i.test(window.navigator.userAgent)
     return true
+    
+  @tagname_is_valid: : (@form_field) ->
```
tjschuck: What's up with this syntax here?  Seems like there's an extra `:` + space.

Additionally, this shouldn't even compile correctly -- did you actually compile the source to test this change in a real Chosen instance?

ComeBurguburu: yeah : is a typo
I correct identation (the problems were wrong identations)
I successfully compiled it with npm coffee 1.9.0

coffee/lib/abstract-chosen.coffee
```diff
@@ -281,7 +282,15 @@ class AbstractChosen
       return false if /Mobile/i.test(window.navigator.userAgent)
     return true
 
+  @tagname_is_valid: (@form_field) ->
+    if @form_field.tagName.toUpperCase() != 'SELECT'
+      console?.error 'Chosen error: invalid tagName ' + @form_field.selector + ' is not a select' 
+      return false
+    if @form_field.options.length == 0
+      console?.error 'Chosen error: ' + @form_field.selector + ' has no options\'s childs '
```
tjschuck: Grammar/formatting:  `' has no <option> children'`

coffee/lib/abstract-chosen.coffee
```diff
@@ -281,7 +282,15 @@ class AbstractChosen
...
+      return false
+    return true
+
   @default_multiple_text: "Select Some Options"
   @default_single_text: "Select an Option"
   @default_no_result_text: "No results match"
-
```
tjschuck: There should still be an empty new line at the end of the file.  But it should be empty, i.e. not including the blank two spaces that you had before.

ComeBurguburu: added

tjschuck: You added it back with the two blank spaces -- see [the diff](https://github.com/harvesthq/chosen/pull/2231/files).  There needs to be a new line, no spaces, which should mean nothing showing in the diff for the last line.

pfiller:
Thanks @ComeBurguburu.

I'm not seeing a bug if Chosen is called on a select with no options. Chosen initializes properly and you can still add options dynamically without issue. If you're seeing differently, please share a [fiddle](http://jsfiddle.net) showing the problem.

I'm not convinced that Chosen should need to worry that you're applying it to a select element. If you're using specific enough selectors, this really shouldn't be a common problem and the breakage is going to be obvious.

For those reasons, I'm going to close this.

ComeBurguburu:
The bug is here: 
http://jsfiddle.net/e6ys6grb/1/

I explain, in my application I duplicate a chosen-select when user click on "add" button.
I used a class to change the select style but also as jQuery selector

The bug is due to the use of the option inherit_select_classes

To correct the bug I replace
```````````````````````````````````````````````````js
$(".test").chosen({inherit_select_classes: true,width: '200px'});
```````````````````````````````````````````````````
by
`````````````````````````````````````````````````````````js
$("select.test").chosen({inherit_select_classes: true,width: '200px'});

**Comment by pfiller** [comment]
You could also just apply Chosen after you've appended the clone:

```JavaScript
$(document.body).append($(".test").clone(false))
$(".test").chosen({inherit_select_classes: true,width: '200px'})
```

PR #2254, opened 2015-03-03
Fix code execution vulnerabilities.

### LINKED ISSUE #2132: XSS vulnerability (new in v1.2.0)

adunkman:
I noticed a scripting vulnerability just now in `1.3.0` when rendering results — and upon checking `master`, not only is that one present, but there’s also another when rendering a single select’s selected option. 

Demo of the issue using [chosen@`501568a`](https://github.com/harvesthq/chosen/tree/501568a69376c7df2e8db803006f94ff74bcc9f5) (current master): http://jsfiddle.net/hpv4La79/3/.

Fixes #2132, replaces #2209, #1806. Effectively undoes #1638.

### DIFF
--- a/coffee/lib/abstract-chosen.coffee
+++ b/coffee/lib/abstract-chosen.coffee
@@ -77,7 +77,7 @@ class AbstractChosen
         if data.selected and @is_multiple
           this.choice_build data
         else if data.selected and not @is_multiple
-          this.single_set_selected_text(data.text)
+          this.single_set_selected_text(data.html)
 
     content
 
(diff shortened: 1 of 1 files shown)


pfiller:
Yup - #1638 should not have merged after all. The Ember folks are going to be displeased by this, but it's the right thing to do.

Thanks, @adunkman!

PR #2257, opened 2015-03-04
correct bower link in README.md

ComeBurguburu:
update bower link to last chosen release (the link throw error 404)

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ However, you can specify that you'd like to use [the release ZIP](https://github
 Either install from the command line with
 
 ```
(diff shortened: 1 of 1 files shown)


tjschuck:
Actually, @pfiller just messed up the tag syntax when cutting the release.  I re-tagged it to be `v1.4.0` like it should be, so the links should work now.

But good catch!  Thanks @ComeBurguburu!

pfiller:
I never mess up.

tjschuck:
@pfiller :kissing_heart:

PR #2263, opened 2015-03-11
Use grunt-gh-pages to build our pages

pfiller:
@tjschuck I'm probably gonna merge these changes and release a new version. Because: YOLO!

This removes some old grunt build stuff and replaces it with some new grunt build stuff (that does all the dirty work for us). It also served as a sanity check to see what was working in our gruntfile. Turns out that most of it works as expected.

Here's the new release process (which I'll try out after landing this).

1. Update version number in `package.json`
1. run `grunt prep_release` (updates all of our files except README)
1. Update version numbers in `README` (this step will go away in another PR)
1. New version commit!
1. Add generated .zip file to the release (and delete it from your local)
1. run `grunt gh-pages`

There's a way to automate the version bump, but I think it's kind of stupid. This process isn't very hard and we don't need to do it that often. Heck, just having it documented is a big step up.


pfiller:
I forgot a step between 5 & 6 ... which is to publish the release. Luckily, I didn't forget it in real life.

.gitignore
```diff
@@ -1,10 +1,10 @@
 .DS_Store
 node_modules
 .project
-dist/
```
stof: are you sure about that ?

Gruntfile.coffee
```diff
@@ -117,4 +107,4 @@ This file is generated by `grunt build`, do not edit it by hand.
     json2 = pkg.jqueryJSON
     json1[key] = json2[key] for key of json2
     json1.author.name = pkg.author
-    grunt.file.write('chosen.jquery.json', JSON.stringify(json1, null, 2))
+    grunt.file.write('chosen.jquery.json', JSON.stringify(json1, null, 2) + "\n")
```
stof: We could drop this json file entirely given that the jQuery plugin registry is abandonned: http://plugins.jquery.com/

koenpunt: :+1:

.gitignore
tjschuck: It looks like the updated Gruntfile no longer writes anything to `dist/`, so that's probably why.

pfiller: Yup - that was added with `build-gh-pages` and shouldn't be used any longer. If you have a `dist/` folder left in your repo, you can just trash it.

Gruntfile.coffee
pfiller: Will deleting the package file remove it from the read-only list they have there? Should we even care?

I used the jQuery file for writing a lot of the Bower stuff in #2264. Removing it would need to be done in that PR (or after it lands).

stof: @pfiller it won't remove it from the list AFAIK (they were parsing this file for tags). Thus, it would mean that the list is not really read-only :smile:.
And even if it removes it, I don't think we care. The jQuery plugin registry never got the traction intended (probably because it waited mor than 1 year before replacing the previous registry so everyone found another solution)

pfiller: Good enough for me. Whenever we decide to move forward on the other PR, I'll make the update to kill this file.

Issue #2261, opened 2015-03-08
Add dist folder


chassq:
Hey Guys,

Great lib! Would you mind adding a dist folder with compiled code? It would be nice to have when we use bower to install pkg. Right now we only seem to get the source and would have to compile ourselves. Would be a nice convenience!

Thanks!

adunkman:
@chassq check out the [releases](https://github.com/harvesthq/chosen/releases) section for compiled code. We’re not so happy to include build artifacts in the source repository, since, well, they’re build artifacts. 

Thanks for the interest, though!

chassq:
The issue is your releases section is not set up for a bower pull. You should consider it. Many other javascipt libraries do this and have a build directory on the main release branch.

Sent from my Windows Phone
________________________________

tjschuck:
Bower does not currently support adding a post-install/build step.  We don't provide compiled sources in the repos for many reasons which have been repeatedly stated throughout the issue tracker.

However, you can specify that you'd like to use [the release ZIP](https://github.com/harvesthq/chosen/releases/latest), which includes the compiled and minified sources.

Either install from the command line with 

```
$ bower install https://github.com/harvesthq/chosen/releases/download/v1.4.0/chosen_v1.4.0.zip
````

or add Chosen to your own project's `bower.json` file, like:

```json
{
  "name": "my-project",
  "version": "1.0.0",
  "dependencies": {
    "jquery": "1.11.0",
    "chosen": "https://github.com/harvesthq/chosen/releases/download/v1.4.0/chosen_v1.4.0.zip"
  }
}
```

See #1509, #1472, and #1333 for more info.

This is even [documented in the README](https://github.com/harvesthq/chosen#bower-installation).

pfiller:
`bower install chosen`

As of 1.4.2, Chosen supports bower installs. Read [the 1.4.2 release notes](https://github.com/harvesthq/chosen/releases/tag/1.4.2) for details.

alvarotrigo:
I personally always find the `dist` folder useful. Even to just compare the amount and size of code I'm going to include in my project or to just compare it with alternatives out there. In this case, `select2`.

stof:
@alvarotrigo having conflicts on all pull requests each time we merge one, because of the dist folder (and so being unable to merge 2 PRs in a row) is a huge burden for maintainers. If you want to see the code size, look at the other repository where we store built files (updated only for releases, not for every push into master)

alvarotrigo:
[...]

Issue #1908, opened 2014-04-24
feature(bower.json): add bower.json


remiwebs:
Bower only recognizes dependencies in bower.json. Because bower.json is missing, dependancies like jquery are not installed. In this case my request is to add bower.json.

Cheers

pfiller:
`bower install chosen`

As of 1.4.2, Chosen supports bower installs. Read [the 1.4.2 release notes](https://github.com/harvesthq/chosen/releases/tag/1.4.2) for details.

Issue #1845, opened 2014-03-20
Add bower ?


doriansavage:

pfiller:
`bower install chosen`

As of 1.4.2, Chosen supports bower installs. Read [the 1.4.2 release notes](https://github.com/harvesthq/chosen/releases/tag/1.4.2) for details.

Issue #1698, opened 2013-12-21
bower install doesn't work


acornejo:
Installing this package through bower does not work.

The reason is that the distributable files (i.e. chosen.jquery.js) is not in the repository.

Please fix this, otherwise people who want to use chosen are forced to install npm, install grunt, and build the package manually. Actually its even worst than that, you also need to install ruby, gem, bundle (I had none of these) to build the css.

Please fix bower install.

pfiller:
`bower install chosen`

As of 1.4.2, Chosen supports bower installs. Read [the 1.4.2 release notes](https://github.com/harvesthq/chosen/releases/tag/1.4.2) for details.

PR #2264, opened 2015-03-11
A Bower Option

pfiller:
@tjschuck @stof @koenpunt 

I don't know about you, but I'm tired of people opening bower issues. I definitely do not want want to add generated js back into our repo, but I figured there had to be _some_ way to do this that would make most people happy.

I thought we could maybe point bower at the generated assets in our `gh-pages` branch, but was disappointed to learn that bower relies on git tags. It seems the only good option is to create a separate repo (as @zurb has done [with Foundation](https://github.com/zurb/bower-foundation)).

So, I've given it a shot. I've created a separate `bower-chosen` repo in my personal account. I didn't want to start it in Harvest's account in case we decide not to go this route. People might actually see it here, but there's little chance of that in my personal account.

This adds two new task:

- `grunt publish-bower`: writes a bower.json file and shoves it (along with the necessary jquery assets) out to the `bower-chosen` repo
- `grunt publish-release`: runs `gh-pages:gh-pages` as well as the `bower-release` task

That means our new deploy process would look like this:

1. Update version number in `package.json`
1. run `grunt prep_release` (builds our assets and generates a new zip)
1. Add generated .zip file to the Github release (and delete it from your local)
1. New version commit!
1. Publish the release (using the Github interface)
1. Run `grunt publish-release`

What do you think? Is this crazy? The only question I have is how this would work with other package managers (some people asked for NPM in #1433). Should we change the name of the other repo and call it `chosen-published`? 

Please share your thoughts!

Addresses #2261, #1845, #1556, #1908, #1641, #1333, #1698, etc. etc. etc.


pfiller:
BTW, you can see what the generated output looks like here:
https://github.com/pfiller/bower-chosen

koenpunt:
What is the reason that we can't push the release packages to the main repo? Because like I see it, the contents of the release package in your bower repo is fine to use as a release package on the main repo.. Or was it that bower depends on the automatically generated tag-package..? Can't remember..

adunkman:
Have you looked into https://github.com/bower/registry/pull/78? I haven’t, but having another repo seems really… gross. I agree it’s less gross than build files in a source repo, but I think this is very clearly a problem with Bower, and seems like something we wouldn’t want to address in Chosen. 

I believe that issue suggests that you can publish versions with something a-like:

```sh
bower register chosen http://harvesthq.github.io/chosen/1.4.0/
```

… but I’m not 100% sure.

adunkman:
Oh, nevermind — seems like that issue supports loading git repositories over http and https, but they’re still required to be git repositories (as per discussion on https://github.com/bower/registry/issues/23).

stof:
@koenpunt the issue with having the compiled files in the main repo is that it leaves us with 2 options:

- updating the dist file in the repo only for releases. Drawback is that all other commits are in an inconsistent state, which is bad (and breaks things for people expecting to get a dev version with bower)
- updating the dist file all the time. this is what we did previously, and it means that all PRs will conflict all the time, making the maintenance a huge nightmare.

@pfiller npm should not be an issue, given that it does not get the code from github (you publish the dist code to the npm registry AFAIK). So npm would not need this special repo.
Note that you should add the Chosen team as collaborators on your repo too, so that other people than you can create releases as well. Having it in the same organization would make it easier by reusing the same Github team instead of duplicating the collaborator management

stof:
@pfiller I think the bower-chosen repo should contain a small Readme file explaining the goal of this repo and pointing to the chosen repo for contributions (and issues should be disabled on the repo)

koenpunt:
@stof I don't mean including them in the repo, just in the published archive ;)

pfiller:
I agree that it's gross to have a separate repo for the generated files, but the dependency on tags doesn't leave us a lot of wiggle room if we decide to support bower. The really nice thing about this is that, once it's set up, this flow doesn't require any extra work on our part. It doesn't seem like the extra repo will need much, if any, maintenance.

> I think the bower-chosen repo should contain a small Readme file explaining the goal of this repo and pointing to the chosen repo for contributions (and issues should be disabled on the repo)

100% yes. A link back to the main Chosen repo should exist anywhere it's possible.

> Note that you should add the Chosen team as collaborators on your repo too, so that other people than you can create releases as well. Having it in the same organization would make it easier by reusing the same Github team instead of duplicating the collaborator management.

Yes! Hadn't thought of that. One thing to note is that you'd need to add a "Personal access token" to your GitHub account for authenticating to the repo via the command line the way it's set up right now. Although, I suppose that might not be necessary if it's in the same organization.

Anywho, it seems like nobody has any blocking feedback on the actual implementation of this. It's really just a question of if you think we should do it or not. Bower generates a lot of noise and I'd like to try and do things to make this repo less noisy. That way, when we're spending time on Chosen, it's on things that actually matter.

A :+1: or :-1: from you guys would be real helpful. If it's +1'ed in concept, I'll go ahead and change the repo to harvest and add y'all.

stof:
@koenpunt the issue is precisely that Bower does not support using the published archives

stof:
:+1: on this concept

Btw, foundation is updating it from Travis-CI for each push to the main repo, so that even dev versions are available. What do you think about this idea (it would also solve the collaborator thing as only Travis would need the access) ?

koenpunt:
:+1: for automated builds on commit, and maybe a travis setup could be a nice reboot of our testing attempts :)

Ninir:
:+1: Yeah, interesting idea. How much work do you guys think you would need to fix this?

pfiller:
Going to merge this and get it working with Travis today!

pfiller:
Published: `bower install chosen`

Ninir:
:+1: Great work guys!

PR #2268, opened 2015-03-15
fix #1562 now the behavior is the same as the proto version

### LINKED ISSUE #1562: When clicking on an option it disappear
I'm having a problem with the latest build from github, not the downloaded version from the site.

In the example file, multi selection dropdown, when holding CTRL and clicking on an option it disappears, but when I open the selection box again the option that I previously selected has the "selected" styled (grey color).

I saw that if I add:

```javascript
high.addClass("result-selected")
```

to the line result_select function, just before the `item = @results_data` it works again.

I don't know if this is the correct behavior or just a bug.

Thanks!

laantorchaweb:
fixed #1562 this class was being added on the prototype version but not in the jquery one. Now is working as expected.

### DIFF
--- a/coffee/chosen.jquery.coffee
+++ b/coffee/chosen.jquery.coffee
@@ -345,6 +345,8 @@ class Chosen extends AbstractChosen
       else
         this.reset_single_select_options()
 
+      high.addClass("result-selected")
+
       item = @results_data[ high[0].getAttribute("data-option-array-index") ]
       item.selected = true
 


pfiller:
Thanks @laantorchaweb. I just looked through the code a bit and I think that we should actually _remove_ the line from the Prototype version as it's no longer necessary. The only thing that's using the class is `winnow_results_highlight` and it's called _only_ after `winnow_results` adds the class name.

Going to open a quick PR to do removal.

pfiller:
I'm wrong. It is needed... just saw #1562.

Going to merge this!

pfiller:
Thanks again, @laantorchaweb!!

PR #2270, opened 2015-03-17
Changed <b> to <stong>

nonumber:
To comply with accessibility standards.


tjschuck:
There's nothing inherently non-accessible about `<b>` tags, so calling this an "accessibility" issue seems inaccurate.

I believe what you're referring to is that `strong` is more _semantic_ than `b` — that is to say "this is strongly emphasized text", not just "make it bold".

_However_, here that's not true.  The group label _is not emphasized_.  It's merely visually distinct.  `<b>` is a [valid HTML5 tag](http://www.w3.org/TR/html-markup/b.html), indicating:

> a span of text offset from its surrounding content without conveying any extra emphasis or importance

which is exactly how it's being used here.

Though you could make an argument that this should instead be something like a `span` with a class that is then styled as bold, changing it to `strong` is _not_ the correct solution.

See also [this article](http://html5doctor.com/i-b-em-strong-element/) about `i`, `b`, `em`, and `strong` in HTML 4 and 5.

PR #2281, opened 2015-03-26
Remove unused line

pfiller:
@tjschuck @stof @koenpunt 

A small thing. While investigating #2268, I noticed that this line wasn't necessary. This is me removing it!

Thanks to @laantorchaweb for finding it!


pfiller:
Whoops - I did something bad here.

pfiller:
See #2282

PR #2283, opened 2015-03-27
Get a base travis file up and atom

pfiller:
@harvesthq/chosen-developers 

I like the Travis idea proposed in #2264 – so let's do it. This is a first step that doesn't do much.


pfiller:
The light is green, the trap is clean.

PR #2285, opened 2015-03-28
Use bundler to install gems

pfiller:
@adunkman 

We're using more than just Compass from our Gemfile. Should have used bundler from the start.


adunkman:
Yay! +1

Issue #2288, opened 2015-03-28
Bower improvements


hostep:
Hi

I just noticed you added support for bower, which is great!
But I have a few remarks:
- You should also add the prototype build, not only the jquery build
- You should also add the non-minified versions (both js and css)

Thanks!

pfiller:
I agree with the non-minified versions, but am not sold on shipping the Prototype version. jQuery is the top library on Bower's install stats and, as far as I can tell, there isn't even an official bower package for Prototype.

hostep:
Ah, well let me explain.
We use the Chosen library in a few Magento projects, and Magento ships with Prototype by default, so we already have Prototype in our project, it isn't pulled in with bower.

SimoSca:
Command
````
$ bower install chosen
````

does install a wrong version. After time, i've noted the `bower-chosen` repo.

I suggest update the [chosen readme](https://github.com/harvesthq/chosen) for other user.

I've solved with

````
$ bower install harvesthq/bower-chosen --save
````

Bye!

stof:
@pfiller please register the ``chosen`` package in the Bower registry

jblotus:
is there any way we can have unminified  versions of js and css included in bower-chosen ? we have our own minification pipeline and prefer unminified for development

pfiller:
Going to close this up. The non-minified versions are now in bower and the repo seems to have been properly registered in the bower repository.

PR #2298, opened 2015-04-01
Correct css file used in bower main declaration

calumbrodie:
When installed via bower, the file 'chosen.css' is not installed, only 'chosen.min.css'. This change means that the main declaration doesn't need to be overridden manually in bower.json.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
   "bowerJSON":{
     "main": [
       "chosen.jquery.min.js",
-      "chosen.css",
+      "chosen.min.css",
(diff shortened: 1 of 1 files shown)


calumbrodie:
relevant to #2288

shicholas:
:+1:, ran into the same issue.

pfiller:
I just took a look through some common JavaScript libraries and it looks like including the unminified versions in Bower is standard operating procedure. Instead of changing the CSS to include the minified version, I think we should change the JS to include the unminified version! 

Thanks anyways @calumbrodie

PR #2155, opened 2014-12-03
Add dppx to Retina media query

peterjmag:
This addresses #1771, which is now fixable because the Chrome team has 
fixed the erroneous "Consider using dppx" console message.

See also:
https://code.google.com/p/chromium/issues/detail?id=336276
http://css-tricks.com/forums/topic/chrome-console-recommends-dppx-instead-of-dpi-even-if-dppx-is-used/#post-187595

PR #2304, opened 2015-04-06
Use ‘dppx’ units instead of ‘dpi’

ElijahLynn:
Getting this error in Chrome:

"Consider using ‘dppx’ units instead of ‘dpi’, as in CSS ‘dpi’ means dots-per-CSS-inch, not dots-per-physical-inch, so does not correspond to the actual ‘dpi’ of a screen. In media query expression: (-webkit-min-device-pixel-ratio: 1.5), not all, (min-resolution: 144dpi)"

selection_285

### DIFF
--- a/sass/chosen.scss
+++ b/sass/chosen.scss
@@ -417,7 +417,7 @@ $chosen-sprite-retina: image-url('chosen-sprite@2x.png') !default;
 /* @end */
 
 /* @group Retina compatibility */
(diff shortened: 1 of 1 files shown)


ElijahLynn:
Cross-referencing a downstream project this is used on https://www.drupal.org/node/2466181

tjschuck:
This was fixed in #2155, which was shipped with [version 1.4.0](https://github.com/harvesthq/chosen/releases/tag/v1.4.0) of Chosen.

ElijahLynn:
Ahh, thanks.

PR #2306, opened 2015-04-07
Update bower-chosen to use unminified versions

pfiller:
@harvesthq/chosen-developers 

Though Bower itself is unopinionated about whcih version is included, the consensus amongst common libraries is to include the unminified versions of source files. People probably use build systems to minify those files for production (I hope).

Fixes #2298 and #2288

### DIFF
--- a/bower-publish.sh
+++ b/bower-publish.sh
@@ -12,7 +12,7 @@ git config --global user.name "bower-chosen"
 
 git clone https://pfiller:${GH_TOKEN}@github.com/harvesthq/bower-chosen.git
 rm -rf bower-chosen/*
-cp public/bower.json public/*.png public/chosen.jquery.min.js public/chosen.min.css bower-chosen/
+cp public/bower.json public/*.png public/chosen.jquery.js public/chosen.css bower-chosen/
 cd bower-chosen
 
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -37,7 +37,7 @@
   ],
   "bowerJSON":{
     "main": [
-      "chosen.jquery.min.js",
+      "chosen.jquery.js",
       "chosen.css",
```
pfiller: No need to change the css since it was already referencing the full version (even though we were packaging the minified css file).

bower-publish.sh
```diff
@@ -12,7 +12,7 @@ git config --global user.name "bower-chosen"
 
 git clone https://pfiller:${GH_TOKEN}@github.com/harvesthq/bower-chosen.git
 rm -rf bower-chosen/*
-cp public/bower.json public/*.png public/chosen.jquery.min.js public/chosen.min.css bower-chosen/
+cp public/bower.json public/*.png public/chosen.jquery.js public/chosen.css bower-chosen/
```
pfiller: GitHub wasn't very consistent with period selection:
weird, wild stuff

koenpunt:
Shouldn't both the minified and unminified versions be in there?

Issue #155, opened 2011-08-04
Dropdown at the bottom of the page breaks layout


JAStanton:
If you have dropdown that's at the bottom of your page. The hidden dropdown menu is pushing the bottom of the page down.  That needs to be hidden so that it doesn't effect the styles of the page, it would be nice if the drop down detected the bottom of the page, did a little math to figure out if it has room to expand or not and if it doesn't have room it will switch styles to appearing above and now bellow.

raysunqi:
Why there is no dynamic solutions after all the years? I like the control very much except this problem.

mveldhuizen:
Thanks @fenisteel. Your solution works great and very clear instructions on what needed to be changed.

stof:
@fenisteel can you send a PR with your changes ?

rockerest:
+1 to a PR from @fenisteel. I use Chosen via Bower in production, and can't edit library files in the repository.

totszwai:
+1 wish this could get out asap in v1.3.1

lode:
I've turned the changes by @fenisteel into a PR, see https://github.com/harvesthq/chosen/pull/2312.

peimn:
<p>This works for me</p>
<pre><code>jQuery('.chosen-single').click(function() {
        var $chosePosition = this.getBoundingClientRect(),
            $choseDropListHeight = $(this).next().outerHeight();
        if($(this).hasClass('above')) $(this).removeClass('above');
        if (($chosePosition.bottom + $choseDropListHeight + 10) > $(window).height() && ($chosePosition.top - $choseDropListHeight) > 10) {
            $(this).next().addClass("above");
            $(this).addClass("above");
        }
        else {
            $(this).next().removeClass("above");
            $(this).removeClass("above");
        }
        $chosensingle = $(this);
        $chosenvar = this;
        $(window).bind('mousewheel DOMMouseScroll', function(event){
            $chosePosition = $chosenvar.getBoundingClientRect(),
                $choseDropListHeight = $chosensingle.next().outerHeight();
            console.log($chosePosition.bottom + $choseDropListHeight + 10, '-',$(window).height(), '-',$chosePosition.top - $choseDropListHeight);
            if (($chosePosition.bottom + $choseDropListHeight + 10) > $(window).height() && ($chosePosition.top - $choseDropListHeight) > 10) {
                $chosensingle.next().addClass("above");
                $chosensingle.addClass("above");

            }
            else {
                $chosensingle.next().removeClass("above");
[...]

PR #2036, opened 2014-08-10
Add screen edge detection

craigmdennis:
- Using information from pull request #1330 as well as in the issue conversation.
- Ignored `.map` files.

Based on code by @MattDease and @fenisteel

PR #1330, opened 2013-07-04
Add viewport edge detection to results dropdown

MattDease:
Fixes harvesthq/chosen#155

Ensure dropdowns are always visible in the viewport by checking if they would drop below the window's bottom edge and positioning them above the input if that is the case.
This code is based off the snippets by @coreyworrell

This addition is especially important when a select input is positioned near the bottom of the page where the user can't scroll anymore - without this addition the input is nearly unusable as the list of options is cut off.

PR #2312, opened 2015-04-20
Lets the selectbox "drop-up" when it would fall out of the window

lode:
This adds drop-up support to the select box, based on https://github.com/harvesthq/chosen/issues/155. The changes are made by @fenisteel (which based the work on changes by @coreyworrell) and shared in the mentioned issue. I just turned it into a PR as I'd love to see this feature being merged into the project.

I only copied the currently shared code without changes (except the notes, see below). I can imagine it doesn't adhere to the code style guide, please let me know if that is the case.

One thing left open, as [mentioned](https://github.com/harvesthq/chosen/issues/155#issuecomment-48858566) by @abulhasanlakhani, is that this only calculates the drop-down/drop-up on page load. It would be great if that could be improved, but otherwise, I'd say it is already an improvement as it is.

### DIFF
--- a/coffee/chosen.jquery.coffee
+++ b/coffee/chosen.jquery.coffee
@@ -169,8 +169,8 @@ class Chosen extends AbstractChosen
     @container.addClass "chosen-container-active"
     @active_field = true
 
-    windowHeight = $(window).height() + $('html').scrollTop(),
-    totalHeight  = @dropdown.height() + Math.ceil(@dropdown.offset().top);
+    windowHeight = $(window).height() + $('html').scrollTop()
+    totalHeight  = @dropdown.height() + Math.ceil(@dropdown.offset().top)
 
     if (totalHeight > windowHeight) {
       @container.addClass "chosen-with-dropup"
@@ -244,8 +244,8 @@ class Chosen extends AbstractChosen
     @container.addClass "chosen-with-drop"
     @results_showing = true
 
-    windowHeight = $(window).height() + $('html').scrollTop(),
-    totalHeight  = @dropdown.height() + Math.ceil(@dropdown.offset().top);
+    windowHeight = $(window).height() + $('html').scrollTop()
+    totalHeight  = @dropdown.height() + Math.ceil(@dropdown.offset().top)
 
     if (totalHeight > windowHeight) {
       @container.addClass "chosen-with-dropup"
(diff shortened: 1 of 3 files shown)


tjschuck:
@lode [Check the Travis output for your commit](https://travis-ci.org/harvesthq/chosen/builds/59204422) -- your CoffeeScript appears to be invalid and doesn't compile.  (In particular, you don't need to declare variables with `var` in CS.)

digitaljohn:
Any movement on this @lode ?

parterburn:
+1 - is the fix coming that would make the tests pass?

lode:
Coffee is not my field as you can see by the last commits :)
It seems its compiling now, but the tests fail on `$(window)` not being available in the headless browser (I guess), it gives: `TypeError: 'undefined' is not a function (evaluating 'window.height()')`

How to work around that?
What I can find is a quite large fix: http://stackoverflow.com/questions/9821166/error-accessing-jquerywindow-height-in-jasmine-while-running-tests-in-maven

coffee/chosen.jquery.coffee
```diff
@@ -168,6 +169,13 @@ class Chosen extends AbstractChosen
     @container.addClass "chosen-container-active"
     @active_field = true
 
+    windowHeight = $ window.height() + $ 'html'.scrollTop()
```
tomitm: @lode You should include the brackets in this case. `window.height()` is being used instead of `$(window).height()` hence the error in Travis.
```coffee
windowHeight = $(window).height() + $('html').scrollTop()
```

You're most likely getting the issue you mentioned in your comment because this actually compiles to $() enclosing the rest of the line, not just up to the dot, as I imagine you're expecting.

tomitm:
Looking forward to this getting fixed!

@lode Check out http://js2.coffee/ if you want some practice with coffeescript and seeing how it compiles to Javascript. Helps a lot when you're not used to it.

lode:
Thanks @tomitm! I actually used js2.coffee and got that output, but modified it because it didn't seem to comprehend with other parts of the code... Anyway, now it builds.

Is there anyway to get a build version to test it it actually works? (I don't have the build setup locally.)

pfiller:
There are 3 open PRs for this functionality (this one, #2036, and #1330). Of these, #2036 seems the most complete so I'm going to close in favor of this. If I'm missing something, please feel free to shout at me!

Thanks @lode. Very sorry it took us so long to review!!

PR #2319, opened 2015-04-29
Update title and classnames (when option enabled) from select on update

### LINKED ISSUE #2232: chosen: updated does not pick up css.
TL;DR:
As can be seen in this fiddle: http://jsfiddle.net/73qm0whz/

Issue:
chosen: update does not pick up on changed css-classes. If a class has been added dynamically via JS and you after that do the handy $("#idtag").trigger("chosen: updated"); command, Chosen will not pick up the new class. It will pick up on changed options, placeholders and whatnots, but the css-class is never read.

### LINKED ISSUE #2311: Update trigger should update title attribute from original select
We have a select that's being used as a filter, with options populated from an API call.  Sometimes, however, there are no options, and the select is disabled and the title is changed so the user gets a message about WHY it's disabled.   Unfortunately, while chosen does handle the disabled state quite well, it doesn't update itself with the new title that's been set on the original select element.  It should be fairly trivial to accomplish this - I'll look into submitting a pull request with a fix if I find it's a simple fix.

isochronous:
This Fixes #2232 and Fixes #2311.
I simply extracted the logic that determines which classnames to apply to the container into a separate method, did the same for the logic that copies the title attribute, then replaced the original logic with calls to those two methods and added the same calls to the `results_build` method in the appropriate places so that classes and the title will be updated from the select on update.

I also significantly reorganized the spec files to make them more granular.  I originally had the whole virtual DOM being reconstructed and torn down before and after each test, but ran into some issues in prototype when using that approach that I was unable to figure out.  Since none of the current tests (or the tests I added) actually conflict with each other, I just organized the tests so that the virtual DOM and the chosen instance are only created once before any test and then destroyed after all tests.

Both prototype and jQuery versions were updated, and all tests are passing.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -11,6 +11,7 @@ GEM
 
 PLATFORMS
   ruby
+  x86-mingw32
 
 DEPENDENCIES
   compass
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -11,7 +11,6 @@ GEM
 
 PLATFORMS
   ruby
-  x86-mingw32
 
 DEPENDENCIES
   compass
--- a/coffee/chosen.jquery.coffee
+++ b/coffee/chosen.jquery.coffee
@@ -25,15 +25,19 @@ class Chosen extends AbstractChosen
     @is_rtl = @form_field_jq.hasClass "chosen-rtl"
 
   set_up_html: ->
+    container_classes = ["chosen-container"]
+    container_classes.push "chosen-container-" + (if @is_multiple then "multi" else "single")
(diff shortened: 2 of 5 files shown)


onepiece2015:
hello

isochronous:
Hi there

isochronous:
In case anyone wants these versions of chosen before this pull request is accepted (assuming it is), you can find both the jQuery and prototype versions here: https://gist.github.com/isochronous/541fa49aa34f7f5b20a3

isochronous:
Is anyone paying attention to these pull requests?

PR #2320, opened 2015-05-01
Fixed background colours

drmonkeyninja:
Background colours were getting overridden by the background mixin resulting in a redundant background rule in the generated CSS.


pfiller:
Thanks @drmonkeyninja and sorry for the sloooowwww review. It looks like this may have been useful, but ultimately #2479 is going to rewrite the CSS this fixes. I'm going to close this in favor of that PR.

Thanks again!

drmonkeyninja:
No problem.

PR #2360, opened 2015-07-06
close item on touchend event

kirillstepkin:


pfiller:
This PR is dramatically lacking in explanation. What is this solving? Why do we want to add it?

Closing for the time being.

Issue #2359, opened 2015-07-03
Add option for minimum term length to start search


panmari:
I have a select field that has very many options (with large descriptions). When only entering one character, the search lags noticeably, leading to a bad user experience. 

This could be improved by only starting the search after entering n characters of the search string. Is there already such an option in chosen? If not, would it be worth implementing?

tjschuck:
Related:

* https://github.com/harvesthq/chosen/pull/1225
* https://github.com/harvesthq/chosen/pull/1109

Another option instead of waiting for _n_ characters would be to add some kind of debounce functionality to not trigger the filtering until the user has stopped typing instead of on immediately upon every keypress.

panmari:
My pull request doesn't show any results at all until the minimum number of characters are entered for search. It would possibly be preferable to either

a) Show a hint that there are more characters required
b) Show all results (unfiltered)

Unfortunately I'm not familiar enough with the codebase to implement these features. But the tests I provided can be easily adapted to test such behavior.

tjschuck:
Closed by #2362

panmari:
Solved with alternative approach, as described in #2362

PR #2361, opened 2015-07-07
Only start searching after entering specified minimum number of characters.

panmari:
Before reaching this minimum number of characters, no results are shown.
This should improve performance for very large selects that tend to lag
when only entering very few characters.
Default behavior is to start immediately searching, as before.


panmari:
See #2359, code inspired by #1225.

Issue #2363, opened 2015-07-09
Deprecated linear-gradient syntax


altras:
Howdy!

In our project we're using https://github.com/jenius/autoprefixer-stylus and chosen.js and from that combo I saw a couple of warnings like:
```
{ type: 'warning',
  text: 'Gradient has outdated direction syntax. New syntax is like "to left" instead of "right".',
  node: 
   { type: 'decl',
     parent: 
      { nodes: [Object],
        type: 'rule',
        parent: [Object],
        source: [Object],
        before: '',
        between: '',
        selector: '.chosen-container-single .chosen-single',
        semicolon: false,
        after: '',
        lastEach: 10,
        indexes: {},
        _autoprefixerDisabled: false,
        _autoprefixerPrefix: false },
     source: { start: [Object], input: [Object], end: [Object] },
     before: '',
     prop: 'background',
     between: ':',
     value: 'linear-gradient(top,#fff 20%,#f6f6f6 50%,#eee 52%,#f4f4f4 100%)',
     _autoprefixerDisabled: false,
     _autoprefixerPrefix: false,
     _autoprefixerValues: { '-webkit-': '-webkit-linear-gradient(top, #fff 20%, #f6f6f6 50%, #eee 52%, #f4f4f4 100%)' } },
  plugin: 'autoprefixer' } 0 [ { type: 'warning',
    text: 'Gradient has outdated direction syntax. New syntax is like "to left" instead of "right".',
    node: 
     { type: 'decl',
       parent: [Object],
       source: [Object],
       before: '',
       prop: 'background',
       between: ':',
       value: 'linear-gradient(top,#fff 20%,#f6f6f6 50%,#eee 52%,#f4f4f4 100%)',
       _autoprefixerDisabled: false,
       _autoprefixerPrefix: false,
       _autoprefixerValues: [Object] },
    plugin: 'autoprefixer' } ]
```

I see in some places (https://github.com/harvesthq/chosen/blob/master/sass/chosen.scss#L278, https://github.com/harvesthq/chosen/blob/master/sass/chosen.scss#L67) that indeed chosenjs is using deprecated linear-gradient syntax `top` instead of `to top` 

Since it's an easy fix  I would like to understand in the first place is it necessary to use the old deprecated syntax :? Seeing this post https://github.com/postcss/autoprefixer/issues/473#issuecomment-119508310 -> `Wrong gradient direction creates real problems.` I'm wondering how we should approach this warning? xD

[...]

PR #2364, opened 2015-07-10
Remove "top" from linear-gradient - it's the wrong syntax

mlettini:
@tjschuck @pfiller @koenpunt @altras

Fixes https://github.com/harvesthq/chosen/issues/2363

Our linear gradients go from top-to-bottom, and according to the [latest syntax](http://www.w3.org/TR/css3-images/#linear-gradients) we should be using `to bottom`, but since that's the default we don't need to specify that. Having no direction will go `to bottom` anyway.

### DIFF
--- a/sass/chosen.scss
+++ b/sass/chosen.scss
@@ -34,7 +34,7 @@ $chosen-sprite-retina: image-url('chosen-sprite@2x.png') !default;
   a{
     cursor: pointer;
   }
-  
+
   .search-choice, .chosen-single{
     .group-name{
       margin-right: 4px;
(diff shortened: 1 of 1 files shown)


tjschuck:
@mlettini Does this have any browser-support issues?  I.e. was the old syntax better supported in older browsers?  Or does everything that supports the old also support the new?

mlettini:
@tjschuck This should not have any browser-support issues. Also there's 4 places where we're currently using `linear-gradient` with no direction anyway: https://github.com/harvesthq/chosen/blob/master/sass/chosen.scss#L180,
https://github.com/harvesthq/chosen/blob/master/sass/chosen.scss#L216,
https://github.com/harvesthq/chosen/blob/master/sass/chosen.scss#L249, https://github.com/harvesthq/chosen/blob/master/sass/chosen.scss#L313

tjschuck:
@mlettini Seems legit.

Dunno if you want to get a +1 from @pfiller (or someone else from @harvesthq/chosen-developers and/or @harvesthq/designers?), but feel free to :shipit: if you don't think you need one.

Issue #2372, opened 2015-07-16
Trigger event when search results are updated


iamcal:
It would be useful to be able to hook into the updating of search results, by triggering a `chosen:filter` or similar event when `Chosen.prototype.update_results_content()` is called. This would allow mutation of the results HTML during search.

PreVaDu:
If someone needs the same, just edit prototype function like this way:
```
    Chosen.prototype.update_results_content = function(content) {
      res = this.search_results.html(content);
      this.form_field_jq.trigger("chosen:showing_results", {
        chosen: this
      });
      return res;
    };
```

billyshena:
Hi, I have tried to edit prototype function inside the chosen.jquery.js file but still does not work.
Is there any updates so far for this issue?

Best,

PreVaDu:
Have you already make function `chosen:showing_results`?
In my case, I have something like this:
`$('.chosen-select').on("chosen:showing_results", function(){
...
});`

atreyamisra:
I found that if I change the function Chosen.prototype.no_results_clear to the following, this is achieved @PreVaDu:
Chosen.prototype.no_results_clear = function() {
      this.search_results.find(".no-results").remove();
      return this.form_field_jq.trigger("chosen:showing_results", {
        chosen: this
      });
    };

PR #2373, opened 2015-07-17
trigger a `chosen:filter` event when the results list changes

iamcal:
this event allows decoration/mutation of the results html, as used by https://github.com/websemantics/Image-Select

implements #2372


websemantics:
Is there an update on this issue?

pfiller:
If this was going to get merged, it'd need to happen in both the Prototype and jQuery versions. At the moment, there are two other PRs which do this (#1626 and #2356) so i'll close in favor of one of those.

Thanks anyways, @iamcal

PR #2382, opened 2015-07-30
Wraps label's for attribute with double quote.

klebervirgilio:
Hi,

Today I faced a problem while trying to apply chosen to some HTML inputs with `'` _(single quote)_ into its `id`s.

Besides the fact that isn't appropriate to use single quotes into HTML's `id` attributes, was wondering, if perhaps that small contribution could be accepted.

Thanks.

### DIFF
--- a/coffee/chosen.jquery.coffee
+++ b/coffee/chosen.jquery.coffee
@@ -263,7 +263,7 @@ class Chosen extends AbstractChosen
   set_label_behavior: ->
     @form_field_label = @form_field_jq.parents("label") # first check for a parent label
     if not @form_field_label.length and @form_field.id.length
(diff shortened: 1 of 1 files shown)


pfiller:
Oh gosh you sent me down the rabbit hole on this one. It turns out _both_ quotes and single quotes are allowed in IDs these days. Back in my day IDs had all kinds of rules! Anywho...

Given that things will break for some type of quotation mark (whether it's ' or "), I'm inclined to maintain the status quo. I prefer the single quotes to escaped double quotes. I know that's just one guys opinion, but that's how it goes sometimes!

Thanks for sending me reading all kinds of weird things, @klebervirgilio!

PR #692, opened 2012-06-20
Add functionality to allow templates

cjstewart88:
I've added the functionality to allow templates in dropdowns by using the `data` attribute on `options` and a `template` option when setting up chosen.

I've compiled the code and provided jquery and prototype examples.

Below is the jQuery example.

### DOM:

    <select data-placeholder="Choose your favorite candy..." class="chzn-select-template-example" style="width:350px;" tabindex="11">
      <option value=""></option>
      <option data-type="chocolate" data-user-rating="1.8">M&Ms</option>
      <option data-type="fruity" data-user-rating="1">Starburst</option>
      <option data-type="fruity" data-user-rating="7">Skittles</option>
      <option data-type="chocolate" data-user-rating="9">Reese's</option>
      <option data-type="chocolate" data-user-rating="10">Snickers</option>
    </select>

### Initialize It:

    $(".chzn-select-template-example").chosen({
      template: function (text, templateData) {
        return [
          "<span style='float: right'>User Rating: " + templateData.userRating + "</span>",
          "<div>" + text + "</div>",
          "<div><i>type of candy: " + templateData.type + "</i></div>"         
        ].join("");
      }
    });

### Result:

image

Thanks to Elliott Wood for code reviewing and to Select2 for inspiration.

PR #2405, opened 2015-08-15
Add ability to set images for select options

banesto:
If following markup:
```
<select data-placeholder="Choose a country..." class="chosen-select-images" style="width:350px;" tabindex="19">
  <option data-img-src="https://cdnjs.cloudflare.com/ajax/libs/flag-icon-css/0.7.1/flags/4x3/be.svg">Belgium</option>
  <option data-img-src="https://cdnjs.cloudflare.com/ajax/libs/flag-icon-css/0.7.1/flags/4x3/lv.svg" selected>Latvia</option>
  <option data-img-src="https://cdnjs.cloudflare.com/ajax/libs/flag-icon-css/0.7.1/flags/4x3/us.svg">USA</option>
  <option data-img-src="https://cdnjs.cloudflare.com/ajax/libs/flag-icon-css/0.7.1/flags/4x3/jp.svg">Japan</option>
  <option>Neverland</option>
</select>
```
and setup is in place:
```
$(".chosen-select").chosen({option_images: true});
```

it's possible to display images next to options in results dropdown and selected choices:
chosen-options-image

### DIFF
--- a/coffee/chosen.jquery.coffee
+++ b/coffee/chosen.jquery.coffee
@@ -29,6 +29,7 @@ class Chosen extends AbstractChosen
     container_classes.push "chosen-container-" + (if @is_multiple then "multi" else "single")
     container_classes.push @form_field.className if @inherit_select_classes && @form_field.className
     container_classes.push "chosen-rtl" if @is_rtl
+    container_classes.push "chosen-image-container" if @option_images
 
     container_props =
       'class': container_classes.join ' '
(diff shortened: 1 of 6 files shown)


tjschuck:
See also: https://github.com/harvesthq/chosen/pull/692

banesto:
Should it mean that it will not be merged for at least 3 forthcoming years or what? :smile:

tjschuck:
Not necessarily, but the biggest blocker to something like this getting merged is certainly much more philosophical than just the code.

Chosen is intended to be a drop-in replacement for a native `select`, with an eye towards progressive enhancement.  Native `select`s have no such templating functionality, so this is (perhaps) a step too far outside the goals of the project.

That's not to say it would _never_ be added, but features that break the bounds of native `select`s functionality are always a difficult consideration.

For a different philosophical take, you can consider [Select2](https://select2.github.io), which was originally a fork of Chosen.  Select2 _does_ offer templating, as well as countless other features that have been considered for Chosen but ultimately decided against.  Select2 seeks to be much more configurable, but potentially at the cost of complexity.

banesto:
Ok, that's what I thought. Thank you for explanation.

banesto:
I was thinking also about setting results via ajax and that's clearly out of chosen scope. As for image display in options, I consider it to be a tiny visual tweek for original html's select and that could possibly be in the scope.

tjschuck:
@banesto Heh, may I point you to https://github.com/harvesthq/chosen/pull/166?

pfiller:

Sorry this has sat open so long @banesto. The hope is that by getting things under control, we can get to a place in the future where we can once again start adding new functionality (at which point we'd come back to closed PRs like this). Thanks for your patience and understanding!

PR #2407, opened 2015-08-15
Add native placeholder option

banesto:
Main benefit is that placeholder is shown when search input is in focus but there's nothing entered yet, so in case of multiple selects when there is at least one option already selected search input will stil show what information user should input:
default text as placeholder


pfiller:

Sorry this has sat open so long @banesto. The hope is that by getting things under control, we can get to a place in the future where we can once again start adding new functionality (at which point we'd come back to closed PRs like this). Thanks for your patience and understanding!

pfiller:
#2200 is similar

PR #2435, opened 2015-09-30
Remove mystery asterisk

tjschuck:
@pfiller You added this asterisk in https://github.com/harvesthq/chosen/commit/067e6b98cfd35153e519de2d89eb7792c7a62b4c

I cannot figure out why it's there.

Because it seems to be doing nothing but confusing _very important people like me_, I'm removing it.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -1331,7 +1331,7 @@
         <ul>
           <li>
             <h3>Form Field Change</h3>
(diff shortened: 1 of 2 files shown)


tjschuck:
@stof My man :heart: :heart: :heart_eyes_cat: :heart:

pfiller:
I loved that asterisk.

here's another

koenpunt:
While at it, I notice on that same line this: _That let's you do something like this:_
Where I'm quite sure it should be _lets_.

Issue #2424, opened 2015-09-20
chosen select does not work in tables when dynamically creating rows


KillSwitch666:
can't access dynamically created select menus.
using latest JQuery , FF 40.0.3, windows 7 Ultimate 64bit.

PS.
using: $("[class^='chosen_select_']").chosen({

UPDATE:
created a fiddle
https://jsfiddle.net/vwgyt42u/

mav2287:
I dealt with this same issue the other day. The problem is that when you pull the html you are pulling the select AND the chosen box the was created for it. Now you have identical copies and that is bad. What you want to do is just copy the select and then apply chosen to it to make it a chosen drop. If you really want to use HTML you would need to strip out the chosen box and then re-apply chosen to it.

tjschuck:
> If you really want to use HTML you would need to strip out the chosen box and then re-apply chosen to it.

Note that there is a `$("#my-select").chosen("destroy")` to strip Chosen, but it's undocumented -- see #2290 :sob: (PR to document it gladly accepted!).

After that, you can re-initialize with `$("#my-select").chosen()`.

Closing per @mav2287's workaround -- thanks @mav2287!

mav2287:
tjschuck thanks for the heads up on the chosen destroy!  Please do add that to the docs I can think of at least 4-5 places in my code alone I could have used it rather than a workaround.

tjschuck:
@mav2287 See #2436!

KillSwitch666:
@mav2287 thanks for the comment i already solved it by using destroy as mentioned by @tjschuck, it just drove me mad until i figured it out.
i didn't had the chance to update it here because the project i was working on got delayed because of this and i had to work extra hours and fast, so i forget all about this post.
anyway here is an updated fiddle with destroy method implemented: https://jsfiddle.net/vwgyt42u/1/

stojko75:
It's worth mentioning, you need to destroy it **before** row cloning and **after** cloning, initialize again.

Rig-ved:
Thanks for the info shared .. Need to destroy the chosen and then initialize :) worked wonders

sudhir600:
Here is the full working code - (Gist) [dynamically_chosen.html](https://gist.github.com/sudhir600/a5e5c69097187eb576c54d714d63c717) and [JsFiddle ](https://jsfiddle.net/sudhir600/2f71xmLz/)

Issue #2290, opened 2015-03-29
Is there any way to unmount chosen ?


nicolas-besnard:
I can't find it in the documentation. I'd like to unbind `chosen` from my `<select>`

koenpunt:
Easy
```js
$('select').chosen('destroy');
```

Or use `enable` / `disable`

koenpunt:
But good that you bring this up, because apparently this is nowhere to be found in the documentation..

koenpunt:
Oh my bad there's no enable or disable, but destroy should work.
@harvesthq/chosen-developers anyone feels up to the task of adding docs for destroy? Or am I missing something?

tjschuck:
@koenpunt Added in #2436 (finally!).

PR #2436, opened 2015-09-30
Add docs for destroying Chosen

tjschuck:
As discussed on https://github.com/harvesthq/chosen/issues/2290 and https://github.com/harvesthq/chosen/issues/2424#issuecomment-144467347, this adds docs for destroying Chosen.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -1326,7 +1326,7 @@
         </div>
       </div>
 
-      <h2><a name="change-update-events" class="anchor" href="#change-update-events">Change / Update Events</a></h2>
+      <h2><a name="change-update-events" class="anchor" href="#change-update-events">Change / Update / Destroy Events</a></h2>
       <div class="side-by-side clearfix">
         <ul>
           <li>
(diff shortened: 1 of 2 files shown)


koenpunt:
Great start, but destroy isn't a event, it's a method ;)

koenpunt:
The options page is probably more suited for documenting destroy?

tjschuck:
> The options page is probably more suited for documenting destroy?

Two problems with that:

1. It's not an option (as you pointed out in your first comment!  It's a method).  Same reason initialization isn't documented on the options page either.

2. There's only _one_ options page, and the docs for destroy are different for the jQuery and Prototype versions :-\

koenpunt:
Hm yeah, and `destroy` is the only method.. if we're strict we should create a separate section for "methods".

koenpunt:
But 
> It's not an option (as you pointed out in your first comment! It's a method).

The events are also documented on this page.. So I guess we're not that strict anyway :sob:

mav2287:
In my mind it would make the most sense to put this right under Custom Width Support on the main project page with all the other examples of usage. You could even do a simple demo with a button labeled destroy to illustrate it.

tjschuck:
@koenpunt The header for that section was already a little odd, since one event was triggered and the other was triggerable.  I changed the header to hopefully better lump the three things into a more obvious section of "lifecycle" concerns.

It's not great, but barring a complete rewrite/restructing of the docs, I think it's okay...

Eh?

koenpunt:
:+1:

Issue #2439, opened 2015-10-05
Chosen is enabled on Blackberry browser but dropdowns are not usable


tremby:
On the Blackberry Z10 device I'm testing with, Chosen is enabled. It looks like the [code only checks for Iphone etc al and Android](https://github.com/harvesthq/chosen/blob/4b03346635d59ca6028ca267590b84354b8eead4/coffee/lib/abstract-chosen.coffee#L284). But the dropdown won't scroll on touch events and so Chosen is somewhat unusable.

Is the lack of a test for Blackberry intentional, or could it be added?

This device identifies itself as `Mozilla/5.0 (BB10; Touch) AppleWebKit/537.35+ (KHTML, like Gecko) Version/10.2.1.3247 Mobile Safari/537.35+`.

tjschuck:
@tremby Blocking Blackberry is probably a good idea.  Submit a PR if you're able!

tremby:
Well there are similar things in #1848 and #1850 which haven't been merged. I'd probably just do `/Mobi/.test(navigator.userAgent)` as [suggested by MDN](https://developer.mozilla.org/en-US/docs/Browser_detection_using_the_user_agent#Mobile.2C_Tablet_or_Desktop). I'll submit a PR for that if it would be an acceptable solution? According to MDN you then wouldn't need the iPhone/iPod/Android tests, and you probably wouldn't need the mobile clause of the IE test either.

tjschuck:
`Mobi` is no good because that includes tablets, particularly iPads -- see https://github.com/harvesthq/chosen/pull/1850 that you linked above for that.

https://github.com/harvesthq/chosen/issues/1848 also looks good (and similar to this), but that was also just an issue, not a PR, so couldn't be merged :)

tremby:
Ah, the meaning of the comment where that was mentioned wasn't clear to me. Noted. But as I've just posted in #1850 I'm not going to just copy and paste someone else's code and commit it with my own name on it, so I'll wait for its author to respond. Closing this in favour of that.

Issue #1848, opened 2014-03-20
Extended Mobile Browser detection


StephKoenig:
I added support  for detecting Windows Mobile, older IEMobile, BB10 and older BlackBerrys on my local version of 1.1.0 using the following code snippet - I had to move the IE detection to the bottom, hope this helps

```Javascript
AbstractChosen.browser_is_supported = function() {
  if (/iP(od|hone)/i.test(navigator.userAgent)) {
    return false;
  }
  if (/Android/i.test(navigator.userAgent)) {
    if (/Mobile/i.test(navigator.userAgent)) {
      return false;
    }
  }
  if (/IEMobile/i.test(navigator.userAgent)) {
          return false;
  }
  if (/Windows Phone/i.test(navigator.userAgent)) {
          return false;
  }
  if (/BlackBerry/i.test(navigator.userAgent)) {
    return false;
  }
  if (/BB10/i.test(navigator.userAgent)) {
    return false;
  }
  if (window.navigator.appName === "Microsoft Internet Explorer") {
    return document.documentMode >= 8;
  }
  return true;
};
```

Also we found in our testing of multiple phones and OSs that not all properties of the window object are available but accessing properties of navigator works fine which is odd because they're the same (or should be) and window.navigator is more precise/explicit...

tremby:
@StephKoenig, see https://github.com/harvesthq/chosen/issues/2439. I don't want to just copy and paste your code for my own PR; maybe you can?

tjschuck:
@tremby I appreciate your respect for authorship here!

Just two notes, though, since this is an old issue, and @StephKoenig might have moved on:

1. Per our contributing document, submitted patches are also covered by our MIT license, so you can freely use the code here if you'd like.

2. However, if you'd like, you can still give @StephKoenig authorship!  When you commit, just add the flag `--author="Stephane Koenig <stephane.koenig@gmail.com>"` to your commit, and he will be given authorship on the patch.  It will look like [this commit](https://github.com/harvesthq/chosen/commit/d73404a19ce0d0517cbe4648edf5d08afeb43511) -- see how one user "authored" but another "committed"?  You'd be the "committed" there.

Thanks again for trying to be nice about it! :heart: :heart: :heart_eyes_cat: :heart:

tremby:
I've had one too many patches get accepted but tweaked slightly by the maintainer and in the process my name stripped out of them to assume anything at all. But yes, in this case this seems like an excellent solution.

[...]

PR #2440, opened 2015-10-05
Extend mobile browser detection

tremby:
On top of the existing detections of Android and IOS, detect Windows Phone, IE Mobile, Blackberry and BB10.

As discussed in https://github.com/harvesthq/chosen/issues/1848 and https://github.com/harvesthq/chosen/issues/2439

### DIFF
--- a/coffee/lib/abstract-chosen.coffee
+++ b/coffee/lib/abstract-chosen.coffee
@@ -282,12 +282,20 @@ class AbstractChosen
   # class methods and variables ============================================================
 
   @browser_is_supported: ->
-    if window.navigator.appName == "Microsoft Internet Explorer"
-      return document.documentMode >= 8
     if /iP(od|hone)/i.test(window.navigator.userAgent)
       return false
     if /Android/i.test(window.navigator.userAgent)
       return false if /Mobile/i.test(window.navigator.userAgent)
+    if /IEMobile/i.test(window.navigator.userAgent)
+      return false
+    if /Windows Phone/i.test(window.navigator.userAgent)
+      return false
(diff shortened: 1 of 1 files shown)


tremby:
One thing I question is whether the tests should really be case-insensitive. I would lean towards them not being. Maybe it doesn't make much difference.

coffee/lib/abstract-chosen.coffee
```diff
@@ -282,12 +282,20 @@ class AbstractChosen
   # class methods and variables ============================================================
 
   @browser_is_supported: ->
-    if window.navigator.appName == "Microsoft Internet Explorer"
-      return document.documentMode >= 8
     if /iP(od|hone)/i.test(window.navigator.userAgent)
       return false
     if /Android/i.test(window.navigator.userAgent)
       return false if /Mobile/i.test(window.navigator.userAgent)
+    if /IEMobile/i.test(navigator.userAgent)
```
tjschuck: Looks like you're missing a `window.` on all of these conditions.

tremby: It would still work, would it not? window being the global object and all. But I'll add it for consistency.

tremby: Replaced the commit.

tjschuck:
This looks good to me, but hopefully we can get a :+1: from a more qualified @harvesthq/chosen-developers -- maybe @koenpunt if we're lucky?

koenpunt:
I'm not aware of the actual user agent strings, but this seems good to me

tjschuck:
Thanks @tremby and @StephKoenig (and @koenpunt)!

PR #2446, opened 2015-10-09
Fixes #2095

### LINKED ISSUE #2095: $elem.chosen('destroy') initializes chosen
If you perform a chosen('destroy') on an element for which chosen has not yet been initialized, the code does the opposite and initializes chosen on that element. I would expect that in such a case rather nothing would happen.

	$.fn.extend({
			 chosen: function(options) {
		  if (!AbstractChosen.browser_is_supported()) {
			return this;
		  }
		  return this.each(function(input_field) {
			var $this, chosen;
			$this = $(this);
			chosen = $this.data('chosen');
			if (options === 'destroy' && chosen) { // *)
			  chosen.destroy();
			} else if (!chosen) {
			  $this.data('chosen', new Chosen(this, options));
			}
		  });
		}
	});

*) For elements not initialized with chosen, the variable `chosen` will be `false`, hence the whole condition is false, bringing the execution to the next if. And because said variable `chosen` is false that if statement's block is executed, performing `new Chosen(this, options)`, and by that initializing chosen.

yallups:
If option passed is 'destroy' then only call the #destroy method and terminate.
This fixes the issue where a new chosen instance is created unintentionally.

### DIFF
--- a/coffee/chosen.jquery.coffee
+++ b/coffee/chosen.jquery.coffee
@@ -8,9 +8,11 @@ $.fn.extend({
     this.each (input_field) ->
       $this = $ this
       chosen = $this.data('chosen')
-      if options is 'destroy' && chosen instanceof Chosen
-        chosen.destroy()
-      else unless chosen instanceof Chosen
+      if options is 'destroy'
+        if chosen instanceof Chosen
+          chosen.destroy()
+        return
+      unless chosen instanceof Chosen
         $this.data('chosen', new Chosen(this, options))
 
       return


tjschuck:
@yallups Can you squash this down to one commit?  [Instructions](http://blog.steveklabnik.com/posts/2012-11-08-how-to-squash-commits-in-a-github-pull-request).

tjschuck:
This seems okay to me as far as patching the current implementation goes.

@koenpunt You self-assigned #2095 a while ago -- you want to :+1: / :-1: this?

yallups:
Welp. I messed that up somehow. let me try again.

yallups:
Squashed! Thanks @tjschuck for the help.

koenpunt:
It should function like this, so I'm fine with the merge. But I'm still not happy with how the `if/else/return` looks. Also the duplicate conditions. But I'm also not sure what would be the best logic.
 But that could be a later change.

Thanks @yallups!

koenpunt:
And a small note on the commit message. The textual "subject" is often of more value than the issue id.

So where you now have:

> Fixes #2095
If option passed is 'destroy' then only call the #destroy method and terminate.

You're better of the other way around:

> If option passed is 'destroy' then only call the #destroy method and terminate.
Fixes #2095

yallups:
Cool. Great note. thanks for that

tjschuck:
Thanks @yallups and @koenpunt :heart: :heart: :heart_eyes_cat: :heart: !

Draft release notes for the next version have been updated accordingly.

koenpunt:
For example, bootstrap uses the following plugin initializer:

```js

  function Plugin(option) {
    return this.each(function () {
      var $this = $(this)
      var data  = $this.data('bs.alert')

      if (!data) $this.data('bs.alert', (data = new Alert(this)))
      if (typeof option == 'string') data[option].call($this)
    })
  }
```

They initialize if not yet initialized, and after that they call the method. 

So for Chosen that would mean that calling `chosen('destroy')` on a non Chosen instance would create one and immediately destroys it.

Not sure if that's the way to go either..

Issue #2324, opened 2015-05-08
1.4.2 bower.json 


mromanuk:
The version in `bower.json` is incorrect, please also review the main section

`chosen.css` should be `chosen.min.css` 
```json
  "main": [
    "chosen.jquery.min.js",
    "chosen.css",
    "chosen-sprite@2x.png",
    "chosen-sprite.png"
  ],
```
These are the files included with a `chosen.min.css`
```
total 104
-rw-r--r--  1 martin  staff   1164 Mar 28 01:02 bower.json
-rw-r--r--  1 martin  staff    538 Mar 28 01:02 chosen-sprite.png
-rw-r--r--  1 martin  staff    738 Mar 28 01:02 chosen-sprite@2x.png
-rw-r--r--  1 martin  staff  27969 Mar 28 01:02 chosen.jquery.min.js
-rw-r--r--  1 martin  staff  11115 Mar 28 01:02 chosen.min.css
```

wefarrell:
Seeing this issue as well.

stof:
the 1.4.2 tag is indeed messed up in the bower-chosen repo.

@pfiller looks like the wrong commit was tagged. you may want to check the release script

Bishbulb:
This is causing us problems as well. Can a hotfix be published with just this fix?

dannykopping:
Which commit should it tagged against?

lambert-velir:
+1

raywang945:
Causing problems as well

robertjlooby:
^^ still messed up

tjschuck:
@pfiller FWIW, I think we should just release a 1.4.3 version that is fixed, and not try to mutate 1.4.2.  Does bower have any ability to "yank" bad versions?

thisischrisj:
+1

marcelaerts:
+1

vjanssens:
Quick fix for 1.4.2. Add the 'overrides' object to your bower.json file to override Chosen's main object:

````json
{
  "dependencies": {
    "jquery": "~2.1.1",
    "chosen": "~1.4.2"
  },
  "overrides": {
    "chosen": {
      "main": [
        "chosen.jquery.min.js",
        "chosen.min.css"
      ]
    }
  }
}
````

JonCatmull:
Same issue please fix

sylvainbaronnet:
+1

a-k-g:
+1

ashokfernandez:
[...]

PR #2449, opened 2015-10-11
Bower json updates

### LINKED ISSUE #2448: Bower configuration is not following spec
Some things I noted:

- `"main"` includes images, which it should not.

  > Do not include assets files like images, fonts, audio, or video

- `"version"` is deprecated

  > Use git or svn tags instead. This field is ignored by Bower.

https://github.com/bower/spec/blob/master/json.md

koenpunt:
I've updated the bower.json to follow the spec.

Also, `bower-publish.sh` now compares the git tag of `chosen` and `bower-chosen` to see if it needs an update, instead of grep'ing the bower.json

fixes #2448


bower-publish.sh
```diff
@@ -7,6 +9,8 @@ if [ $CURRENT_BRANCH != 'master' ] ; then
   exit 0
 fi
 
+CHOSEN_VERSION=`git tag -l "*.*.*" | sed 's/^v//' | sort -V | tail -1`
```
tjschuck: So I know this script is intended to be run on Travis (which by default is Ubuntu), but is this script ever intended to be run locally and/or anywhere else?  `sort -V` is not supported on OS X (or any BSD).

Worth double checking that it's even supported on Travis?  Their [environment docs](http://docs.travis-ci.com/user/ci-environment/) don't mention anything about coreutils.

bower-publish.sh
```diff
@@ -15,13 +19,13 @@ rm -rf bower-chosen/*
...
+LATEST_VERSION=`git tag -l "*.*.*" | sed 's/^v//' | sort -V | tail -1`
 
-if [ -z $LATEST_VERSION ] ; then
+if [ "$LATEST_VERSION" = "$CHOSEN_VERSION" ] ; then
   echo "No Chosen version change. Skipped tagging"
 else
-  echo "Chosen version changed. Tagging version ${LATEST_VERSION}\n"
-  git tag -a "v${LATEST_VERSION}" -m "Version ${LATEST_VERSION}"
+  echo "Chosen version changed. Tagging version ${CHOSEN_VERSION}\n"
+  git tag -a "${CHOSEN_VERSION}" -m "Version ${CHOSEN_VERSION}"
```
tjschuck: Is this the cause of https://github.com/harvesthq/chosen/issues/2324?  If so, this fixes that too!

bower-publish.sh
koenpunt: `git tag` has support for sorting, but due to the inconsistency in tags (with and without v prefix) that isn't working out either.. (cc #2450)

```
$ git tag --sort=v:refname
0.9
0.12.1
0.13.0
0.14.0
1.0.0
1.4.0
1.4.2
semver
v0.9
v0.9.1
v0.9.2
v0.9.3
v0.9.4
v0.9.5
v0.9.6
v0.9.7
v0.9.8
v0.9.9
v0.9.10
v0.9.11
v0.9.12
v0.9.13
v0.9.14
v0.9.15
v0.10.0
v0.10.1
v0.11.1
v0.12.0
v1.0.0
v1.1.0
v1.1.1
v1.1.2
v1.2.0
v1.3.0
v1.4.0
v1.4.1
```

bower-publish.sh
koenpunt: I believe not, this only makes sure the tag matches the main repo's tag.

But I believe #2324 is not really relevant anymore, because this script has been updated before.

Although a new release hasn't been published to the bower repo since...

bower-publish.sh
koenpunt: But we could retag `1.4.2` as `v1.4.2`, which solves this problem

koenpunt: Or skip `v1.4.2`, and just publish the next release :sparkles:

bower-publish.sh
tjschuck: @koenpunt I think after your handful of pending changes get merged, we should release a new version.

Check out the draft release notes here: https://github.com/harvesthq/chosen/releases -- since at least #2440 and #2446 are both _not_ backwards compatible, I think semver dictates that we should jump to 2.0.0 for the next release.  What do you think?  (At the _very_ least, minor-version bump to 1.5.0.  This is not a teeny-version bump.)

koenpunt: #2446 is actually a bugfix, the previous behavior was unintended. 

But those mobile browsers.. Yeah probably should be `v2.0.0`. But I think we should a `v1.4.3` along that, to fix the bower issue for `v1.4.2`

tjschuck: What would 1.4.3 be?  Just 1.4.2 reissued?

koenpunt: Yeah, mostly for the bower repo

koenpunt:
I've updated with the `git tag --sort` thing, assuming we're going with the `v` prefix

koenpunt:
I'm not sure if this is still going to work on because as far as I can see tags are not build by Travis.. 

For a private project I'm currently using a basic webhook implementation with a small nodejs application on Heroku, which runs a script on the creation of tags.

So If this doesn't work out with travis we could do that as well.

stof:
@koenpunt AFAIK, tags are built by Travis (at least, they were until recently)

koenpunt:
@stof according to travis-ci/travis-ci#1532 they do, but I just tried it and it didn't work..

pfiller:
Thanks for this @koenpunt. Sorry it took 4 months to get it merged. Sigh.

koenpunt:
I'm glad that it works as I intended/expected :sweat_smile:

Issue #2451, opened 2015-10-12
Invalid CSS property declaration at: *


MareoRaft:
Hey guys,

I am not a SASS guru, but I think there is a small error in the SASS / scss file around the very beginning:

    /* @group Base */
    .chosen-container {
      position: relative;
      display: inline-block;
      vertical-align: middle;
      font-size: 13px;
      zoom: 1;
      *display: inline;

Perhaps it is supposed to be `* display: inline;` (with a space), or maybe `&* display: inline;`.  Anyway, when compass converts it to css, it remains as `*display: inline;` and produces the warning "Invalid CSS property declaration at: *" in Safari.  So it seems that the browser ignores that property altogether.

Thanks!
-Matt

MrGeneric:
Hey Matt, it looks like a CSS hack to only apply that line to an old version of IE.

tjschuck:
That's my guess as well, but cc: @mlettini.

mlettini:
Correct. That + `zoom: 1` are hacks for IE. I didn't even remember that was in there, but I'm almost positive it is no longer needed, as the only browsers that would render that CSS are IE6&7, which Chosen doesn't support. Every other modern browser would ignore that line completely. I can probably remove that line, but don't have an IE8 (or 6 or 7 just to test the breakage) instance on-hand at the moment to verify. Might be able to get to it later this week unless someone wants to beat me to it.

PR #2455, opened 2015-10-13
Remove zoom:1 and *display:inline

mlettini:
@tjschuck @koenpunt @pfiller Fixes https://github.com/harvesthq/chosen/issues/2451#issuecomment-147406851

These two properties are hacks that were needed for versions of IE before 8. Chosen now currently doesn't even render for versions of IE before 8, so they are no longer needed. This also fixes Invalid Property error, as described in the Issue above.


koenpunt:
Looks fine to me :+1:

PR #2463, opened 2015-10-21
fixed CSS file name in generated bower.json

SKoschnicke:
The file in the generated bower repository is named `chosen.min.css`, but the generated bower.json references a (not existing) `chosen.css`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
   "bowerJSON":{
     "main": [
       "chosen.jquery.js",
-      "chosen.css",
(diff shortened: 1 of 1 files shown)


koenpunt:
This isn't the right solution. Instead the build script should be altered to include the non-minified file.

koenpunt:
And as you can see in `bower-publish.sh`, this is already the case, so the next release will be correct

SKoschnicke:
All right. When will the next working release be available?

Issue #2443, opened 2015-10-07
Does not work on Android tablets (e.g. Nexus 9)


mydea:
I know that there is quite a bit going on around the topic of mobile support. However, I'm not sure if this issue is known or if this is a "new" bug. On my Android phone (Nexus 4), chosen works as expected (= it falls back to a native select, which works fine). However, on my android tablet (Nexus 9), chosen is displayed normally, which would be OK, but it is impossibly to select an option in the list. This means that I can filter normally, and everything works fine, but I cannot select an entry by clicking on it.

tjschuck:
@mydea Here's the rule for disabling Chosen on Android: https://github.com/harvesthq/chosen/blob/adf64cca88a69f0cb5a37d10c90bc32de774e059/coffee/lib/abstract-chosen.coffee#L287-L288

As you can see there, any user agent that has `Android` _but also_ `Mobile` is blocked.

I assume your Nexus 9 is considered "non mobile" by user agent.  Can you send the user agent that your tablet provides?

Anyway, assuming your Nexus 9 is `Android` but not `Mobile`, this seems to be behaving as intended.  However whether that check is _sufficient_ is worth discussing.

I'm going to edit the title of your issue to make it slightly more accurate for what the issue is.

mydea:
Yeah, I agree that it is correct to display chosen on the Nexus 9. However, it should be possible to select options there, shouldn't it?

tjschuck:
> However, it should be possible to select options there, shouldn't it?

Based on the rules, I would hope so!  But I can't confirm or deny (no Android tablet), so your report stands.  If you can figure out a fix, PRs encouraged!

Just to round out your report, can you clarify what browser, browser version, and OS version are you using?  And what version of Chosen?  If not the latest version, does it still happen on the [latest version](https://github.com/harvesthq/chosen/releases)?  Can you confirm that the [Chosen demo page](http://harvesthq.github.io/chosen/) (which is running the latest version) shows the issue?

mydea:
Sure:

Nexus 9
Android 5.1.1
Chrome for Android 45
User Agent: Mozilla/5.0 (Linux; Android 5.1.1; Nexus 9 Build/LMY48M) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/45.0.2454.94 Safari/537.36

I also tried the Chosen demo page, it's the same there: the chosen select apepars, but I cannot select an option.

[...]

PR #2468, opened 2015-11-03
Remove check for "Mobile" on android devices

mydea:
Chosen has some issues on some android devices like the Nexus 9. This commit will deactivate chosen on all android devices.

See Issue #2443

### DIFF
--- a/coffee/lib/abstract-chosen.coffee
+++ b/coffee/lib/abstract-chosen.coffee
@@ -245,7 +245,7 @@ class AbstractChosen
       when 27
         this.results_hide() if @results_showing
         return true
-      when 9, 38, 40, 16, 91, 17
+      when 9, 38, 40, 16, 91, 17, 18
         # don't do anything on these keys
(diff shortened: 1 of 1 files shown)


pfiller:
Thanks @mydea, but this PR is not currently merge-able because it combines 3 separate issues into one:

- **Remove check for "mobile" on Android:** seems like this might be a good idea.
- **Disable Chosen on iPad:** Has there been consensus somewhere on this? I'm not sure I'm 👍  on this change.
- **Don't open up chosen on option keyup:** Chosen isn't currently opening for me on option press. Not sure why this is necessary and we'd want to see a fiddle showing the issue first.

At the very least, these should be in separate PRs. The Android change is the one most likely to get merged given the above info.

mydea:
Hey! So, the 3. part is apparently a leftover upstream commit from harvesthq/master that was later removed again -  It took me a while to figure out how that ended up here ;)

I'll make another PR with just the android changes.

PR #2474, opened 2015-11-10
Add autoprefixer

fuudstamps:
I created this branch for our Rails app because we kept on getting inline messages while running our test suite: ```autoprefixer: /Users/dude/dev/app_name/app/assets/stylesheets/application.scss:8587:3: Gradient has outdated direction syntax. New syntax is like "to left" instead of "right".``` and they got really annoying next to our pretty green dots.

BTW we really love using this library!


tjschuck:
@fuudstamps I believe your issue has already been fixed in https://github.com/harvesthq/chosen/pull/2364/, which has not yet been rolled into [a release version](https://github.com/harvesthq/chosen/releases) of Chosen.

If you need the updates from master sooner, you can download and build the source yourself in the meantime.

koenpunt:
Although your changes are incomplete, I like the grasp of it.
If besides adding auto prefixing, we also add a fallback mixin for `image-url`, we can skip the ruby/compass dependency altogether

PR #2475, opened 2015-11-10
Don't open up chosen on option keyup.

braddunbar:
Hello Chosen!

When embedding chosen in a chrome extension, the keyboard shortcut includes the option key. If a chosen instance happens to be focused at the time, releasing the option key opens it, but shouldn't. This patch ignores option keyup events.

I ran the build task after making the change like it says to in `contributing.md`, but it didn't change any git-tracked files that I could tell. Let me know if I need to do something else!


tjschuck:
:+1: LGTM B-RAD.

tjschuck:
(Draft release notes updated.)

PR #2477, opened 2015-11-16
use ctrl when option is too many

Dwrite:
add a function,if you press ctrl ,you don't need to click chose the option


stof:
Can you describe exactly what the new behavior is ?

pfiller:
It looks like this adds an option that lets a person choose options by simply hovering their mouse after pressing <kbd>ctrl</kbd>. We already have a behavior for that key (it keeps the list open when clicking a selection) and I think having a second behavior is especially confusing. I'm not a fan of adding features that are so undiscoverable.

Thanks anyways, @Dwrite.

PR #2362, opened 2015-07-07
Added option max_shown_results

panmari:
To increase performance of selects with very many options, an option was
added that limits the number of results drawn. By default, this is set
to infinity, so it does not impact the previous behavior if not set.

PR #2481, opened 2015-12-02
Added option to start search after the nth typed character

emamona:
Added integer option min_char_to_start_search to avoid search start on
first keys pressed on keyboard.  This decrease response time for long
lists with common language captions.
Tryed also to handle the backspace, but UI is less intuitive.

### DIFF
--- a/coffee/lib/abstract-chosen.coffee
+++ b/coffee/lib/abstract-chosen.coffee
@@ -31,6 +31,7 @@ class AbstractChosen
     @inherit_select_classes = @options.inherit_select_classes || false
     @display_selected_options = if @options.display_selected_options? then @options.display_selected_options else true
     @display_disabled_options = if @options.display_disabled_options? then @options.display_disabled_options else true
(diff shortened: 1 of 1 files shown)


tjschuck:
We've chosen to use the `max_shown_results` option for the same effect (_i.e._ better performance for long lists).

See https://github.com/harvesthq/chosen/issues/2359 (this same idea) and https://github.com/harvesthq/chosen/pull/2362 (the better alternative) for more info.

This approach (waiting to start search until _n_ chars) is not very user friendly, because it _seems_ as though there are no results when you start typing, even though there are and you just need to type more.

karthikeyar9:
But with the max_shown_results is there a option to scroll down the list and see all options? If the user does not know what to search and wants to see the dropdown

emamona:
Some of my users already pointed on that....
Something like "Next max_shown_results" results on bottom.

karthikeyar9:
That would be great if we can have a button to scroll through all results. Currently with max_shown_results we can only see the number of results as configured initially.

Issue #2368, opened 2015-07-14
Why compass?


whmii:
I'm currently using `chosen-rails` and am unable to update sass to the current version due to the use of compass in chosen. 

I am planning on putting a pr together this evening to remove it, but before I do I wanted to make sure there wasn't a super huge reason that yall were using compass (since it is no longer well maintained and conflicts so directly with current implementations of sass).

:beers:

tjschuck:
/cc @mlettini 

Also, Compass is no longer well-maintained?  Do you have any reference to it being abandoned by its maintainers to back that up?

mlettini:
> Also, Compass is no longer well-maintained? Do you have any reference to it being abandoned by its maintainers to back that up?

Also curious of this.

We are currently using Compass for `image-url` and any `@include`s. It's been in there awhile. While new browsers probably don't need vendor-prefixed gradients and box-sizing, older browsers and support still do. To be honest, I don't know if Chosen has any written browser requirements...

tjschuck:
@mlettini See [this code block](https://github.com/harvesthq/chosen/blob/910a12599c10a11b7005feedb1e53e4b4535100e/coffee/lib/abstract-chosen.coffee#L275-L282) -- anything that passes it is "supported".

The only desktop browser explicitly not supported is IE less-than 8.

koenpunt:
We could switch to node-sass, which uses libsass. Vendor prefixes could be handled by postcss' autoprefixer

whmii:
I think rails-sass would be fine. I'd imagine updating existing tech instead of introducing new ones.

whmii:
Also, I think that prefixing is better relinquished to a tool like Autoprefixer so that things like prefixing requirements do not effect the sass files themselves. this way you can write the code you want to write and the prefixing is done separately.

whmii:
Ah, I missed that yall were using grunt. I have posted a `wip` branch with some changes to the gruntfile. https://github.com/whmii/chosen/tree/wm-sass-rails

It doesn't seem to want to run any grunt commands right now. I'll debug in a bit.

tjschuck:
@whmii Actually, this all starts from this, right?

> am unable to update sass to the current version due to the use of compass in chosen.

Compass and Sass both are just _development_ dependencies for Chosen -- they're just necessary to build the final CSS from the lib's source.  But Chosen's dev dependencies shouldn't affect your ability to run any arbitrary version of Sass and/or Compass.

What problem specifically are you having?

[...]

PR #2483, opened 2015-12-02
use node-sass for building sass

koenpunt:
Replaced compass for node-sass and added autoprefixer for automated vendor prefixes.

This removes the ruby dependency, and makes building way faster.

This also resolves the issues people having when including the Chosen source into their rails projects. 

Related #2368 #2474


koenpunt:
@tjschuck @mlettini @pfiller anyone? (I can't tag @harvesthq/developers anymore)

stof:
this is a great news. I thought about suggesting it the last time I had to build Chosen from source locally

stof:
:+1:

tjschuck:
How will this fit into the Travis => https://github.com/harvesthq/bower-chosen build system?  Seamlessly?

Also, Sass != Compass — were we not using any of the Compass-y bits of Compass?

koenpunt:
> Also, Sass != Compass — were we not using any of the Compass-y bits of Compass?

Just some css mixins which are covered by autoprefixer, and the `image-url()` helper, which is irrelevant because you can already set the url using a variable, so when you want to use the `image-url` helper (in whatever framework) you simply include the sass file like this: 

```scss
$chosen-sprite: image-url('chosen-sprite.png');
$chosen-sprite-retina: image-url('chosen-sprite@2x.png');
@include "chosen";
```

koenpunt:
> How will this fit into the Travis => https://github.com/harvesthq/bower-chosen build system? Seamlessly?

And yes I think it does, ruby dependency is stripped and just a few extra node modules are installed through npm.

koenpunt:
Although I'm not sure if the browser specification `'last 2 versions, IE 8'` covers all we do support, so maybe this should be a little more explicit.

tjschuck:
For the Ruby dependency => pure Node change, I have pretty much no opinion either way, so cool with me.

@mlettini You have any thoughts on the Compass => pure Sass changes?

stof:
@koenpunt I think it is OK for the browser specification

mlettini:
> For the Ruby dependency => pure Node change, I have pretty much no opinion either way, so cool with me.

Same boat. Compass was just a tool to allow us to write a property once and not all of it's prefixes for old browser support. Now, autoprefixer is that tool, and that's fine with me. Saves a dependency. I believe we made this change for Harvestapp internally too.

starzonmyarmz:
I'm excited for this! :+1:

pfiller:
This seems great to me. I compared the css before and after and they seem to be effectively the same. I really don't like `node-sass`'s default nested style for output:

whack

My personal vote would be to change the `outputStyle` setting to `'expanded`, but I'd defer to @mlettini or @starzonmyarmz on that.

@koenpunt do you mind merging master in? It's a pretty easy conflict in `travis.yml`.

koenpunt:
Should the formatting of the resulting CSS matter? Maybe we should go with convention over configuration here. 

I've updated the branch :rocket:

koenpunt:
I'll also squash some of those commits

koenpunt:
While at it, we might consider upgrading our node version, 0.10 is a oldie

.travis.yml
```diff
@@ -3,7 +3,7 @@ sudo: false
 language: node_js
 
 node_js:
-  - "0.10"
+  - "5.9.1"
```
stof: I suggest using ``5.9`` to get the latest patch release automatically

koenpunt:
@stof good point, updated.

starzonmyarmz:
> Should the formatting of the resulting CSS matter?

I think we should either go fully `compressed` or `expanded`. Personally, I like open source projects to have expanded CSS for educational purposes, and have an optional minified version to go along with it. I think only having the expanded version is fine - there's no reason someone using Chosen can't go minify the file themselves.

koenpunt:
@starzonmyarmz the minified version is also generated when building, and the resulting file isn't included in the repo anyway.

pfiller:
The version we're shipping to `bower-chosen` is currently not minified and I don't think it should be. I'd prefer we use `expanded` so we don't make a big diff here:

https://github.com/harvesthq/bower-chosen/blob/master/chosen.css 

Plus, I think the nested thing looks worse!

PR #2489, opened 2015-12-12
Include SASS-Lint. Improve code standarts, code visibility and fix lint issues for chosen.scss

taracila:
I've noticed a lot of issues with code standards in this file.

Initial lint results: https://gist.github.com/taracila/fa241a5e724bf7929bc3
Fixed issues (almost): https://gist.github.com/taracila/60071ce06f490a789d3a

Standarts for SASS code: https://drive.google.com/file/d/0B1AwVhsNLZHYZUo1YVZTby0xUUk/view

Run linter:
run in terminal "scss-lint -c scss-lint.yml"

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,3 +1,4 @@
 source "https://rubygems.org"
 
 gem 'compass'
+gem 'scss-lint', '~> 0.38.0'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,16 +1,36 @@
 GEM
   remote: https://rubygems.org/
   specs:
-    chunky_png (1.2.8)
-    compass (0.12.2)
+    chunky_png (1.3.5)
+    compass (1.0.3)
       chunky_png (~> 1.2)
-      fssm (>= 0.2.7)
-      sass (~> 3.1)
-    fssm (0.2.10)
-    sass (3.2.9)
+      compass-core (~> 1.0.2)
+      compass-import-once (~> 1.0.5)
+      rb-fsevent (>= 0.9.3)
+      rb-inotify (>= 0.9)
+      sass (>= 3.3.13, < 3.5)
+    compass-core (1.0.3)
+      multi_json (~> 1.0)
+      sass (>= 3.3.0, < 3.5)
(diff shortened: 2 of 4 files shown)


tjschuck:
/cc @mlettini, although I'm going to go ahead and guess this has a pretty low likelihood of getting merged.

See the discussion on https://github.com/harvesthq/chosen/pull/2479 about linters/combers/"standards".

mlettini:
Thanks for the effort here @taracila.

As @tjschuck mentioned, there's ongoing work and discussion in https://github.com/harvesthq/chosen/pull/2479. That PR should get most of the way there as this one does, by using more variables and alphabetizing property order (I can also add a pass for better comment formatting). But there will be further discussion on property order and linting and the like after that gets merged. It seems it would be beneficial to implement an automatic way of doing that to the repo.

taracila:
Hi @mlettini and @tjschuck 

This PR is intended for code formatting (it should not have any regression on any types of selects). I'm using this awesome plugin in each project where I have activated select tag and each time I'm required to reformat the chosen.scss because of background images (replaced by font-icons) and poor formatting code.

BTW, an automatic Continuous Integration service can give a results for all issues/errors with js/sass code pushed by each developer so you can see which PR has issues at the code standards layer.

Thanks for checking that and I'm waiting for a better version of SASS/CSS in chosen plugin in next versions.

koenpunt:
Can sass-lint also format the code?
Because if so I guess I prefer using that over CSSComb

taracila:
@koenpunt I'm afraid that it's not. I would not agree to have such automatic code formatter let it be php or css. CSSComb has plugins for various IDE and this feature should be managed at local env. Once the code is pushed, the SCSS-lint would check the code and will throw all issues (if there are any). This is the practice that me and another 50~ team members are following.

koenpunt:
I'm not really sure why this was closed, because, like @taracila said, it can improve code quality and keeps it consistent when integrated with CI. We could also loosen the rules for linting (I assume that's possible?). 
But is this linter also available for node-sass, as we're stripping the ruby dependency altogether in #2483

PR #2495, opened 2016-01-04
use inherit for width when offsetWidth is 0

### LINKED ISSUE #92: if select is initially hidden, doesn't work
This is because when setting the width it will return 0.  Using this technique works: http://www.foliotek.com/devblog/getting-the-width-of-a-hidden-element-with-jquery-using-width/

The code should check if the elmt.is(:visible) and if so use the above technique.  I'm happy to fix in a fork.  @hat is the procedure?  Just patch the coffeescript?  How do you generate js and js min?

### LINKED ISSUE #1272: Bootstrap Modal

### LINKED ISSUE #2309: Select in a div that's hidden on page load
Hi

I seem to have hit an issue using Chosen to generate a select list in a div element that's hidden on page load.  If the div is then shown, the select list has no width and you can't see any item text.  You should be able to see what I mean with the below JSFiddle:

http://jsfiddle.net/mamwuen5/

If the div is *not* hidden on page load, you can hide and show it to your heart's content, as in the below JSFiddle:

http://jsfiddle.net/mamwuen5/2/

Am I doing something wrong, or is this a bug?

Thanks

Chris

koenpunt:
This to accommodate for times the Chosen select is initially hidden, as is the case in bootstrap modals and panels

To verify this, checkout this branch; `inherit-width`, build using `grunt`, open `public/index.html` and run the following in the developer console

``` js
$('.chosen-select').first().chosen('destroy')
  .parent().hide()
  .find('select').chosen()
  .parent().show()
```

Thanks to @pierre-alain-b for this solution!

fixes #92
fixes #1272 
fixes #1329 
fixes #2309 

cc: @pfiller @tjschuck @stof

### DIFF
--- a/coffee/lib/abstract-chosen.coffee
+++ b/coffee/lib/abstract-chosen.coffee
@@ -253,7 +253,9 @@ class AbstractChosen
     setTimeout (=> this.results_search()), 50
 
   container_width: ->
-    return if @options.width? then @options.width else "#{@form_field.offsetWidth}px"
+    return @options.width if @options.width?
+    return "#{@form_field.offsetWidth}px" if @form_field.offsetWidth > 0
+    return "inherit"
 
   include_option_in_results: (option) ->
     return false if @is_multiple and (not @display_selected_options and option.selected)


tjschuck:
@koenpunt Why `inherit` instead of `100%`?

tjschuck:
Specifically, if the parent's width is defined as a percentage, this would inherit the percentage, making Chosen have the width of percentage-of-percentage instead of the actual width of the parent.

koenpunt:
Hm, did not yet try it with a parent which has the width set to a percentage.

AbeyMarquez:
I have a chosen inside a modal. I rendered chosen before the modal was shown. In my case, having 'initial' here causes the chosen to have a smaller width than it should... It doesn't fill a bootstrap column fully. If I do 'auto' instead of 'initial' it works perfectly.

PR #2498, opened 2016-01-07
Remove height clobbering for chosen-choices

bronzehedwick:
This seems like an IE6 hack, but IE6 usage is very low, and it nukes `height` resets with `!important` for every other browser.


tjschuck:
IE6 isn't supported, so if it _is_ an IE6 hack, removing it should be okay.

IE support goes back as far as IE8.  See http://harvesthq.github.io/chosen/#faqs

mlettini:
This does seem like a safe thing to do. I'll also need to make this change in https://github.com/harvesthq/chosen/pull/2479 which I haven't picked back up yet.

bronzehedwick:
Just checking in on this issue - is merging this PR in blocked by #2479 ?

tjschuck:
@bronzehedwick In my opinion, it shouldn't be.  #2479 is effectively setting the stage for a 2.0 release, and this seems like a bug that affects the current versions and should be handled accordingly.

If @mlettini @stof or @koenpunt gives the :+1:, they can go ahead and merge it with my blessing!

Particularly cc: @stof who is usually the expert on our legacy IE support.

koenpunt:
Looks sensible :+1:

tjschuck:
Thanks @bronzehedwick !

bronzehedwick:
:100: Thanks!

PR #2511, opened 2016-01-21
set_selection method for allowing programmatic selection.

airportyh:
This change allows the programmatic selection of the currently selected value for both single and multiple selection. Detailed tests are included. Example:

``` html
<select data-placeholder='Choose a Country...'>
  <option value=''></option>
  <option value='United States'>United States</option>
  <option value='United Kingdom'>United Kingdom</option>
  <option value='Afghanistan'>Afghanistan</option>
</select>
```
``` js
$('select').chosen();
var chosen = $('select').data('chosen');
chosen.set_selection('United States');
$('select').val() // => 'United States'
```

### DIFF
--- a/coffee/chosen.jquery.coffee
+++ b/coffee/chosen.jquery.coffee
@@ -332,6 +332,29 @@ class Chosen extends AbstractChosen
     @current_selectedIndex = @form_field.selectedIndex
     @selected_item.find("abbr").remove()
 
+  set_selection: (value) ->
+    contains = (arr, wanted) ->
+      (item for item in arr when item == wanted).length > 0
+
+    if @is_multiple
+      values = value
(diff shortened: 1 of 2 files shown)


tjschuck:
Chosen already "supports" setting the selection via setting the selection on the underlying `select` and calling `trigger("chosen:updated")`.  This is [documented here](http://harvesthq.github.io/chosen/#change-update-events).

I get that this might be slightly more convenient, but it's adding complexity and code for something that's already not too difficult to do.

Closing this, but let me know if you have a different use case that I'm missing that justifies this.

Thanks!

airportyh:
Ah I see, I must have poorly understood the documentation. Thank you!

tjschuck:
@airportyh No worries!  Thanks again for opening your PR.

PR #2516, opened 2016-01-27
Fixes bug. Prevent touch events propagation.

drola:
I added `evt.preventDefault()` statement to touch events to prevent
triggering events on underlying elements after touch event is
captured in chosen container.


tjschuck:
Is this different than https://github.com/harvesthq/chosen/pull/2119 or https://github.com/harvesthq/chosen/pull/2093, both of which are already in released versions?  What version did you trigger the bug on?  If not the latest version, does it still happen on the [latest version](https://github.com/harvesthq/chosen/releases)?

Additionally, please link to a publicly accessible page (or a [Fiddle](http://jsfiddle.net)) that shows the issue so we can attempt to reproduce.  Alternatively, confirm that the [Chosen demo page](http://harvesthq.github.io/chosen/) shows the issue.

Closing until more info is provided, then we'll reopen.

PR #2523, opened 2016-02-04
removeDiacritics tu support diacritic-less search

### LINKED ISSUE #536: support chars with accents
if you have a item in the dropdown called "testé"
it would be nice to be able to type "teste" to find it.
right now it's not finding it because of the ´ over the e
perhaps chosen could remove all of those accent first?

hanoii:
Attempt to fix #536.

This allows for search with and without diacritics.


pfiller:
Thanks @hanoii.

This is a tough one. There is obviously demand for something like this, but I'm pretty far from convinced maintaining this giant list of character swaps is a good thing for this project. I think I prefer a solution where it's something that extends Chosen vs. being the default for all uses of the library.

Going to close for the time being.

PR #2528, opened 2016-02-08
install latest git to support version sorting

koenpunt:
Followup on #2527 

This enables docker based builds, and updates Git to the latest stable version, to support sorting based on version instead of alphabetically.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,22 @@
+sudo: false
+
 language: node_js
 
 node_js:
   - "0.10"
 
+addons:
+  apt:
+    sources:
+      - git-core
+    packages:
+      - git
+
 before_install:
   - npm install -g grunt-cli
   - gem install bundler
   - bundle install
+
 before_script: grunt build
 
 after_success: ./bower-publish.sh


koenpunt:
But I now see that using Docker doesn't really result in faster build times..

.travis.yml
```diff
@@ -1,12 +1,22 @@
...
+
 language: node_js
 
 node_js:
   - "0.10"
 
+addons:
+  apt:
+    sources:
+      - git-core
```
tjschuck: I'm pretty sure it can just be `git` by itself and not reference `git-core` anymore.  `git-core` was a transitional name because of goofy reasons — it should now just be a dummy package that points to `git` as its only dependency.

koenpunt: `git-core` is the only one [in the whitelist](https://github.com/travis-ci/apt-source-whitelist/blob/dce2f6b8fd2ccaf4fcd41dbe2f17c69e80d6d6a9/ubuntu.json#L123), so maybe it's true for regular apt, but not with the docker packages.

koenpunt: And these are sources, not packages, the package is simply `git`

tjschuck: Interesting.  It looks like that whitelist entry was only added 18 days ago in https://github.com/travis-ci/apt-source-whitelist/pull/225.  This is fine with me then, because I don't feel like PR'ing the addition of `git` to the whitelist :stuck_out_tongue_winking_eye:

pfiller:
Coooooooooool with me. Thanks @koenpunt!

PR #2537, opened 2016-02-15
Automatic scroll to highlighted option

puresamari:
I made the automatic scrolling to the highlighted option an optional parameter with the default state true.

    $('your_select').chosen({
      scroll_to_highlighted: false
    });


koenpunt:
When would it be that scrolling is not desired?

pfiller:
Thanks @puresamari.

Like @koenpunt, I have a hard time imagining the use case for this. Not showing a person the thing they've selected makes this control's ux considerably worse. I'm inclined to close this, but will reconsider if you can provide a solid argument!

puresamari:
Hello @pfiller and @koenpunt.

I am really sorry for my late response but I can quickly explain my problem / the problem i had:
when I implemented Chosen in my project i realised that whenever i am hovering the lower part of the dropdown, chosen automatically scrolls to the active / highlighted item. this is not wanted in my project for design purposes. there is also the problem that scrolls down way to fast if the dropdown is small.

I hope that explains the problem and again im sorry for answering so late :cry:

koenpunt:
It sounds like that it's too specific, how does this feature affect your design?

puresamari:
Thanks for the fast response.

The design has small drop down boxes (about 200px or something) so when you open one of these drop downs and you hover on the bottom 20% of the container it automatically scrolls down because of the highlighting. 
This causes then the mouse to hover the next item which gets highlighted and then scrolls in to view and so on until it is scrolled to the bottom or you leave these bottom 20% with the cursor.

I simply wanted to prevent this by blocking the automatic scrolling.

danieltome:
I think this option is very useful.
We need it in one of our projects, where we need to have the options go up, instead dropping down.
But because they scroll as soon as you're at the bottom, then it's a bit messy.
Here's a quick screencast of what's happening: http://screencast.com/t/QtFiuLbha

PR #2541, opened 2016-02-19
Add option to enable case sensitive search

### LINKED ISSUE #2509: Option for case-sensitive search
Request: add a settings option to make case sensitive searches.

Use case: in my web app I have a drop down with a list of letters: a, b, c, ... A, B, C,... that get used a lot by the content editors. I'd like to speed up their workflow so instead of opening the drop down, pressing z then moving the arrow down (to move from z to Z), they can just press shit-z to get to Z directly.

koenpunt:
Fixes #2509 

Haven't tested this yet though..


tjschuck:
@koenpunt Docs!

pfiller:
Sure, this seems straightforward enough. I agree we should have docs :)

koenpunt:
:sparkles:

tjschuck:
:+1: :+1: :+1:

pfiller:
Just tested this out and agree that it's good to go. Let's get it merged!

Thanks, @koenpunt!!

PR #2545, opened 2016-02-21
Compress PNGs with zopflipng

atdt:
Zopfli is the most efficient DEFLATE compression algorithm, trading run-time
performance for file sizes that are typically 3-8% smaller than those produced
by zlib with the maximum compression setting. Its output is Deflate-compatible,
so no specialized decoder is needed.

This change was created by running zopflipng against all the PNG files in this
repository. The exact invocation was:

```
  git ls-files --exclude-per-directory=.gitignore -- '*.png' \|
    parallel zopflipng -m -y {} {} \;
```


Krinkle:
:+1: 

<http://blog.codinghorror.com/zopfli-optimization-literally-free-bandwidth/>

koenpunt:
Thanks for your contribution, but ImageOptim includes zopfli, and I believe it is already in use. Besides that, @mlettini is doing a mayor overhaul on the css, and the sprite is updated as well, so while at it maybe he could apply the the image compression instead of having it at a separate change.

mlettini:
https://github.com/harvesthq/chosen/commit/dcd6ea87fb34c61fabe1922a824b0dd7ee636392

:+1:

Krinkle:
(which is pull #2479)

pfiller:
Seems like it makes sense to close this and work on getting #2479 merged. Thanks anyways, @atdt.

Issue #2484, opened 2015-12-03
Empty CSS-File in chosen-npm Package


tojak:
The chosen.css-File in the Node-Package "chosen-npm" only contains some comments, but no css!

Shaazaam:
I have noticed this issue as well...it'd be nice to be able to manage this correctly with node.

lucjan:
+1

mmehlhope:
+1

ericrowan:
I really don't understand this. Can someone explain what the issue is here?

tjschuck:
[chosen-npm](https://www.npmjs.com/package/chosen-npm) was not created by anyone on the Chosen team.  You'll have to try to contact the publisher directly and have them update the package to the latest version.

Closing this since it's outside the scope of Chosen itself (until we potentially take over NPM publishing — see #2550).

Issue #1433, opened 2013-07-31
Chosen not in npm?


chenglou:
There's the package.json file, but npm returns no corresponding 'chosen' package.

skysteve:
+1 for browserify support please

jmeas:
The fix here is pretty easy, I think.

1. Add a UMD wrapper
2. Add the built files to the repo
3. Publish the built files to `npm`

If the chosen team is interested in such a PR, let me know and I'll put something together. The built files could also be put on Bower, and then there shouldn't be any problems with developers using Chosen.

tjschuck:
@jmeas 

> Add the built files to the repo

This is not going to happen.  We used to do this, and it was a nightmare for the team.  The compiled JS + CSS will not be added to the repo -- this has been discussed many times on the issue tracker.

Pre-compiled versions are provided via the [releases page](https://github.com/harvesthq/chosen/releases).  If your package manager of choice supports specifying packages via ZIPs, you should use them.  See the [bower instructions](https://github.com/harvesthq/chosen#bower-installation) for an example.

If your package manager of choice does _not_ support ZIPs or using [the GitHub releases API](https://developer.github.com/v3/repos/releases/), you should open an issue with them to add support for this.

jmeas:
Browserify users, you have two choices that I know of. The first is described above, but I'll summarize it once more:

**One: use Bower and the `window`**

Follow the instructions in the README. Unfortunately, you cannot use `npm`, as far as I know, so set up `bower` in your project and install the zip. Add a module to your code with the following stuff:

```js
var $ = require('jquery');

// Chosen requires this :(
window.jQuery = $;

require('path/to/bower_components/chosen_v1.3.0/chosen.jquery');
```

And it works! jQuery will, as you'd expect, be permanently modified, so you won't need to require chosen anywhere else in the app.

I recommend that you ensure that your linter throws an error if you try to access browser globals to ensure that you don't start referencing this `window.jQuery` anywhere else in your code.

===

**Solution two: wrap it yourself**

My apps often have `vendor` directories for the handful of libraries that aren't packaged to handle module loaders. If you add [this UMD wrapper](https://gist.github.com/jmeas/03e0e2479d35e5f42046) around chosen, then you can require it in as usual. This wrapper will also work for AMD users.

===

**Why not just use browserify-shim?**

This is definitely an option. My app's tests run in Node+JSDom, so it needs to work without shimming or aliasing the libraries.

===

[...]

PR #2550, opened 2016-02-24
update packaging, now with npm support

koenpunt:
**This is a work in progress**

- the packaging tasks are now separated from the build tasks
- streamlined generation of manifest files, each of them follow their respective spec
- add npm support for package

Some things that have to be done:

- [x] Verify with users if the package (manifest) contains the necessary files for an npm/bower package
   - Example package: [chosen-package.zip](https://github.com/harvesthq/chosen/files/144374/chosen-package.zip)
- [x] Setup npm publishing (add `api_key` to `.travis.yml`) (more info: https://docs.travis-ci.com/user/deployment/npm)
- [x] Check with current owner of https://npmjs.com/package/chosen-jquery for possible transfer

Optional:
- [x] Rename package repo(?) (cc: @pfiller)
- [ ] ~~Include publishing of docs in travis setup~~
- [ ] ~~Use `TRAVIS_TAG` env var instead of `git tag ...` checking~~

Related #1433 #2484


publish-package.sh
```diff
@@ -5,20 +5,20 @@ set -e
 CURRENT_BRANCH=`git name-rev --name-only HEAD`
 
 if [ $CURRENT_BRANCH != 'master' ] ; then
```
stof: We should also skip it on pull request builds

koenpunt: It has been like this the whole time, so maybe that was an oversight of @pfiller; people can open a pull request from their master branch. Or don't these PR's trigger a build?

tasks/package.coffee
```diff
@@ -0,0 +1,81 @@
...
+      src: ['public/**/*']
+      dest: 'chosen_<%= version_tag %>.zip'
+
+  grunt.config 'gh-pages',
+    options:
+      base: 'public',
+      message: 'Updated to new Chosen version <%= pkg.version %>'
+    src: ['**']
+
+  grunt.registerTask 'package-jquery', 'Update chosen.jquery.json', () ->
```
stof: I suggest deleting ``chosen.jquery.json`` entirely, given that the jquery plugin registry has been shutdown a long time ago now.

koenpunt: Sounds reasonable. But I believe this point has been made before, and had implications. But I can't recall..

publish-package.sh
stof: hmm, actually, the way to detect the current branch should work, as PR builds won't be on the commit on the master branch. You are not using the Travis env variable to detect the branch

koenpunt: Right, if we'd used `TRAVIS_BRANCH` we would have a problem. While looking into the travis environment vars, we might be better of using `TRAVIS_TAG`, because then we don't need the `git tag ..` trickery

stof: Well, this would mean you have to tag locally and push the tag, which is a different release process

publish-package.sh
```diff
@@ -5,20 +5,20 @@ set -e
...
-git config --global user.email "notmyemail@bower-chosen.lol"
-git config --global user.name "bower-chosen"
-
 git clone https://pfiller:${GH_TOKEN}@github.com/harvesthq/bower-chosen.git
 rm -rf bower-chosen/*
-cp public/bower.json public/*.png public/chosen.jquery.js public/chosen.css bower-chosen/
+cp README.md public/*.json public/*.png public/chosen.jquery.js public/chosen.css bower-chosen/
 cd bower-chosen
 
+git config user.email "chosen@harvesthq.com"
```
tjschuck: So the old email address was at least obviously a fake one — this one is also fake, but doesn't look like it.

Do we need an email address at all?

tjschuck:
I'm generally :+1: on the sentiment — one build process, multiple targets.

Just a heads up that @pfiller is on vacation until around March 7th, so you're unlikely to hear from him about any of this until probably a few days after that at best.

sleewoo:
Please consider to update npm package to 1.5.
Thanks.

koenpunt:
@pfiller did you already take a look at this?

publish-package.sh
pfiller: I miss .lol

tasks/package.coffee
pfiller: I'd be fine with killing it.

publish-package.sh
pfiller: That wasn't a helpful comment. I don't remember specifically, but I think email is required.

I actually do prefer an email that is clearly not a real address. chosen@harvesthq.com looks like a real address and people are going to be frustrated when there's nobody there. How about something like chosen.robot@dontemail.me?

pfiller:
I'm also in favor of the general concept of one build process. I still don't know about taking on the NPM responsibility. Bower has proven to be somewhat of a hassle and I don't love that it takes what little time I spend on this project... 

But, I'm willing to talk about it. @koenpunt, it seems like the things that are left might benefit from just working on them together. Want to set up a time to pair on it sometime next week? I think you have my email.

publish-package.sh
divergentdave: I'd recommend chosen@example.com or noreply@harvesthq.com, both read clearly as a nonfunctional email address.

koenpunt: I +1 noreply

tasks/package.coffee
stof: @koenpunt there was an implication before the rebooted jquery plugin registry was killed. But in the meantime, the jQuery fondation decided that they don't need to maintain such registry.

publish-package.sh
tjschuck: `noreply@example.com` !

pfiller: noreply@chosenbuild.mcfakeaddress

tjschuck: `pf@patrickfiller.com` :trollface:

pfiller:
Ok, @koenpunt and I just fought through this together and we feel like it's ready to go!

Thanks a billion @koenpunt!!

Issue #2506, opened 2016-01-17
Scroll + hilight problem with jQuery 2.2


dmifedorenko:
Hello.

Chosen 1.4.2 not work with jQuery 2.2. Check out two examples:
https://jsfiddle.net/oLyfgt7m/1/
https://jsfiddle.net/oLyfgt7m/2/

On both examples try to select item on middle of list:
1. Click on chosen html item.
2. Scroll by mouse on center of list.
3. Move mouse under list.

With jQuery 2.1.4 all work fine, but with 2.2 I cant select right item, scroll of chosen allways go down.

Looks like some bug insade hilight method.

This issue is hight priority for me, I cant upgrade to jQuery 2.2. Thank you.

glukash:
I've noticed exactly the same issue with jQuery 2.2.

Eric-Guo:
Same issue with jQuery 1.12.

alxcube:
hi guys! looks like i've solved this issue
Thereis following code on line 855 in not minified version:

high_top = this.result_highlight.position().top + this.search_results.scrollTop(); 

i've removed scrollTop and it heleped:

high_top = this.result_highlight.position().top;

CSI68:
Hi all !

I've try your solution Alxcube, and it seems working !

Nice work, thx very much !

chrisrauz:
Thanks alxcube. Works perfectly!

eskimoe:
same bug here, it seems they changed a lot with scrollTop() in 2.2 and 1.12, something to do with tests on different browsers..
please provide an official fix :) thx for this great plugin!

dmifedorenko:
@tjschuck is where any chance you will fix this bug soon? I really need jQuery 2.2 in my project and I cant migrate because of this issue.

babelshift:
@dmifedorenko There is a workaround posted above.

dmifedorenko:
@babelshift really? :)
What can I do with this workaround? Patch my local chosen copy? Fork chosen on github and rebase master every time after new chosen release out? Was this workaround tested in all popular browsers?

Chosen is one of about 15 js libraries used in our project. Our company prefer use official releases of minor libraries and not fork or patch when. I will better migrate to select2 for example.

babelshift:
@dmifedorenko Workarounds are a temporary fix until the official repo releases the real thing. Chosen is not a frequently released library. In fact, the last major release was nearly a year ago. As such, you will not be required to rebase and retest for any great amount of time.

[...]

PR #2555, opened 2016-02-29
Fix scrolling issues

german:
Fix scrolling issues: high_top when scrolling long select (with option groups) could grow very fast (~ twice as fast as @result_highlight.position().top)

### DIFF
--- a/coffee/chosen.jquery.coffee
+++ b/coffee/chosen.jquery.coffee
@@ -217,7 +217,7 @@ class Chosen extends AbstractChosen
       visible_top = @search_results.scrollTop()
       visible_bottom = maxHeight + visible_top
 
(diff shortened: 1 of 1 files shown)


tjschuck:
@german Are you using jQuery 1.12.0 or 2.2.0, by chance?  This is actually a bug in jQuery, not Chosen.  See https://github.com/harvesthq/chosen/issues/2506 for more info, but jQuery has released [1.12.1 and 2.2.1](http://blog.jquery.com/2016/02/22/jquery-1-12-1-and-2-2-1-released/) which should fix the issue.

Thanks!

german:
@tjschuck ok, I see now. Yes, you're right. I'm using jQuery 1.12.0. Thank you for pointing this out!

PR #2560, opened 2016-03-02
Update browser support check

koenpunt:
**Old**
```js
  browser_is_supported: function() {
    if (/iP(od|hone)/i.test(window.navigator.userAgent)) {
      return false;
    }
    if (/Android/i.test(window.navigator.userAgent)) {
      if (/Mobile/i.test(window.navigator.userAgent)) {
        return false;
      }
    }
    if (/IEMobile/i.test(window.navigator.userAgent)) {
      return false;
    }
    if (/Windows Phone/i.test(window.navigator.userAgent)) {
      return false;
    }
    if (/BlackBerry/i.test(window.navigator.userAgent)) {
      return false;
    }
    if (/BB10/i.test(window.navigator.userAgent)) {
      return false;
    }
    if (window.navigator.appName === "Microsoft Internet Explorer") {
      return document.documentMode >= 8;
    }
    return true;
  }
```

**New**
```js
  browser_is_supported: function() {
    if ("Microsoft Internet Explorer" === window.navigator.appName) {
      return document.documentMode >= 8;
    }
    if (/iP(od|hone)/i.test(window.navigator.userAgent) || /IEMobile/i.test(window.navigator.userAgent) || /Windows Phone/i.test(window.navigator.userAgent) || /BlackBerry/i.test(window.navigator.userAgent) || /BB10/i.test(window.navigator.userAgent) || /Android.*Mobile/i.test(window.navigator.userAgent)) {
      return false;
    }
    return true;
  }
```

### DIFF
--- a/coffee/lib/abstract-chosen.coffee
+++ b/coffee/lib/abstract-chosen.coffee
@@ -282,20 +282,17 @@ class AbstractChosen
   # class methods and variables ============================================================
 
   @browser_is_supported: ->
-    switch true
-      when /iP(od|hone)/i.test(window.navigator.userAgent)
-         , /IEMobile/i.test(window.navigator.userAgent)
-         , /Windows Phone/i.test(window.navigator.userAgent)
-         , /BlackBerry/i.test(window.navigator.userAgent)
-         , /BB10/i.test(window.navigator.userAgent)
-        false
-      when /Android/i.test(window.navigator.userAgent)
-        !/Mobile/i.test(window.navigator.userAgent)
-      when "Microsoft Internet Explorer" == window.navigator.appName
-        document.documentMode >= 8
-      else
-        true
-
+    if /iP(od|hone)/i.test(window.navigator.userAgent) or
+       /IEMobile/i.test(window.navigator.userAgent) or
+       /Windows Phone/i.test(window.navigator.userAgent) or
+       /BlackBerry/i.test(window.navigator.userAgent) or
+       /BB10/i.test(window.navigator.userAgent)
+      return false
+    if /Android/i.test(window.navigator.userAgent)
+      return false if /Mobile/i.test(window.navigator.userAgent)
+    if "Microsoft Internet Explorer" is window.navigator.appName
+      return document.documentMode >= 8
+    return true
 
   @default_multiple_text: "Select Some Options"
   @default_single_text: "Select an Option"
--- a/coffee/lib/abstract-chosen.coffee
+++ b/coffee/lib/abstract-chosen.coffee
@@ -282,6 +282,8 @@ class AbstractChosen
   # class methods and variables ============================================================
 
   @browser_is_supported: ->
+    if "Microsoft Internet Explorer" is window.navigator.appName
+      return document.documentMode >= 8
     if /iP(od|hone)/i.test(window.navigator.userAgent) or
        /IEMobile/i.test(window.navigator.userAgent) or
        /Windows Phone/i.test(window.navigator.userAgent) or
@@ -289,8 +291,6 @@ class AbstractChosen
        /BB10/i.test(window.navigator.userAgent) or
        /Android.*Mobile/i.test(window.navigator.userAgent)
       return false
-    if "Microsoft Internet Explorer" is window.navigator.appName
-      return document.documentMode >= 8
     return true
 
   @default_multiple_text: "Select Some Options"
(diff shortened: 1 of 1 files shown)


coffee/lib/abstract-chosen.coffee
```diff
@@ -282,21 +282,20 @@ class AbstractChosen
...
-      return document.documentMode >= 8
-    return true
+    switch true
+      when /iP(od|hone)/i.test(window.navigator.userAgent)
+         , /IEMobile/i.test(window.navigator.userAgent)
+         , /Windows Phone/i.test(window.navigator.userAgent)
+         , /BlackBerry/i.test(window.navigator.userAgent)
+         , /BB10/i.test(window.navigator.userAgent)
+        false
+      when /Android/i.test(window.navigator.userAgent)
```
Pezmc: Why not `/Android/i.test(window.navigator.userAgent) and /Mobile/i.test(window.navigator.userAgent);` here?

koenpunt: (some sort of) consistency; this way all tests are on their own line

coffee/lib/abstract-chosen.coffee
```diff
@@ -282,21 +282,20 @@ class AbstractChosen
...
-    if /Windows Phone/i.test(window.navigator.userAgent)
-      return false
-    if /BlackBerry/i.test(window.navigator.userAgent)
-      return false
-    if /BB10/i.test(window.navigator.userAgent)
-      return false
-    if window.navigator.appName is "Microsoft Internet Explorer"
-      return document.documentMode >= 8
-    return true
+    switch true
```
tjschuck: This `true` is unnecessary.

tjschuck:
> even though if/else tends to be faster, the resulting js of this switch is more concise

Should we care about the resulting JS?

coffee/lib/abstract-chosen.coffee
koenpunt: I tried, and CoffeeScript inverses the conditions, but that didn't work out so well for 

```coffee
when "Microsoft Internet Explorer" == window.navigator.appName
  document.documentMode >= 8
```

Because that resulted in:
```js
  case "Microsoft Internet Explorer" !== window.navigator.appName:
    document.documentMode >= 8;

Comment (Pezmc) [inline coffee/lib/abstract-chosen.coffee]: I'm pretty certain it's needed, else it defaults to `switch(false)` and inverses the match.

**Comment by koenpunt** [comment]
I'm actually not sure if `if/else` in JS is faster than a `switch`, but I know that it's the case in some languages.

**Comment by koenpunt** [comment]
And I assumed the resulting JS to be less bytes, but it turns out that's not the case either 🤔 

image

image

**Comment by koenpunt** [comment]
I've reverted so a `if`, with less duplication

**Comment by koenpunt** [comment]
Maybe we should put the `"Microsoft Internet Explorer"` test on top, because regex is more expensive. Although this could also be seen as micro-optimization..

**Comment by stof** [comment]
@koenpunt moving it to the top is a good idea. It may even help UglifyJS optimizations, as it will see a ``ìf (...) {return false } return true`` and be able to shorten it to ``return !(...)``

**Comment by koenpunt** [comment]
image

File: coffee/lib/abstract-chosen.coffee [hunk 3]
```diff
@@ -282,20 +282,15 @@ class AbstractChosen
...
-      return false
-    if window.navigator.appName is "Microsoft Internet Explorer"
+    if "Microsoft Internet Explorer" is window.navigator.appName
       return document.documentMode >= 8
+    if /iP(od|hone)/i.test(window.navigator.userAgent) or
+       /IEMobile/i.test(window.navigator.userAgent) or
+       /Windows Phone/i.test(window.navigator.userAgent) or
+       /BlackBerry/i.test(window.navigator.userAgent) or
+       /BB10/i.test(window.navigator.userAgent) or
+       /Android.*Mobile/i.test(window.navigator.userAgent)
```
tjschuck: @koenpunt There's a subtle change here — this will no longer match the opposite order, _e.g._ `Mobile blah blah Android`, which the original did.

Not sure if anything _has_ a user agent like that, but just pointing out that this is not 100% equivalent.

koenpunt: I know, but `Mobile` always appears _after_ `Android` according to https://developer.chrome.com/multidevice/user-agent

tjschuck: :+1:

pfiller:
I tend to prefer readability over micro optimizations that save us a few characters. In this case, however, it seems we get both! I find the changes slightly more readable so I'm :+1: 

Gonna merge it in. Thanks @koenpunt!

PR #2589, opened 2016-04-07
Added an offset to the calculated width to take account of the deselect button when enabled

dtsn:
I've noticed that when you enable the deselect button the width of the button is not taken into account when calculating the width of the select box. Therefore a lot of the items are trimmed. 

In this pull request i've added an extra 31 pixels to the size of the dropdown. This is the offset in the standard CSS. We could make this dynamic, but I thought this would be a good start.

### DIFF
--- a/coffee/lib/abstract-chosen.coffee
+++ b/coffee/lib/abstract-chosen.coffee
@@ -255,7 +255,7 @@ class AbstractChosen
     setTimeout (=> this.results_search()), 50
 
   container_width: ->
-    return if @options.width? then @options.width else "#{@form_field.offsetWidth}px"
+    return if @options.width? then @options.width else if @options.allow_single_deselect then "#{@form_field.offsetWidth+31}px" else "#{@form_field.offsetWidth}px"
 
   include_option_in_results: (option) ->
     return false if @is_multiple and (not @display_selected_options and option.selected)


mlettini:
This is a JS change, which is not in my wheelhouse. So I leave a review to @koenpunt or @pfiller.

koenpunt:
Thanks. For starters I would like to see this `31` in a variable somewhere, because it now is sort of ambiguous.

dtsn:
@koenpunt I complete agree, we can calculate the computed margin-right of the `.chosen-single span`. However calculating this would have to be after the render and therefore you would get a jump.

It might be better to have it as an option as part of the `allow_single_deselect` setup, defaulting to 31 because thats the default CSS.

koenpunt:
I wasn't talking about calculation, but if you just define it in a variable and use that variable instead of the plain `31`.
Something like `single_select_right_margin = 31`

pfiller:
So, I think this solution is problematic and my preference is not to merge it.

1. The automatic width calculation already has enough issues (just search "width" in our issues) that I'm hesitant to add anything that makes it more complicated -- especially when that complication comes in the form of a magic value (variable or not).
2. People who set the width of their select via CSS _and_ who use the deselect option are going to be confused when Chosen keeps insisting that it needs to be 31px wider.

The easiest fix is to specify your select's width when using deselect. It's an imperfect situation all around and I think that means erring on the side of "less complicated" is probably the best answer.

Thanks anyways, @dtsn.

dtsn:
Maybe it would be a good idea to update the documentation to include the following: 

> If you enable deselect make sure you add the width of the deselect to the width of the textbox.

PR #2590, opened 2016-04-07
Fix scroll issue

Dorious:
### Summary
I removed scrollTop added to the result element top. It was making the result list scrolling crazy and going to the bottom.

Replicate bug here:
https://jsfiddle.net/7L9w067o/1/

I checked and it exists in Chrome & Firefox (both windows & mac)
With removed scrollTop everything is fine.

PS. This exists since 1.0 so I dunno why anybody spot that :P

### DIFF
--- a/coffee/chosen.jquery.coffee
+++ b/coffee/chosen.jquery.coffee
@@ -217,7 +217,7 @@ class Chosen extends AbstractChosen
       visible_top = @search_results.scrollTop()
       visible_bottom = maxHeight + visible_top
 
-      high_top = @result_highlight.position().top + @search_results.scrollTop()
+      high_top = @result_highlight.position().top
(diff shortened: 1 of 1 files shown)


tjschuck:
@Dorious Are you using jQuery 1.12.0 or 2.2.0, by chance?  This is actually a bug in jQuery, not Chosen.  See https://github.com/harvesthq/chosen/issues/2506 for more info, but jQuery has released [1.12.1 and 2.2.1](http://blog.jquery.com/2016/02/22/jquery-1-12-1-and-2-2-1-released/) which should fix the issue.

Closing for now in case that's your issue, but if I'm wrong, just let us know and we'll reopen.

tjschuck:
@Dorious In fact, I thought your fix looked familiar — it was the exact same one as #2555, so it's _probably_ the same issue.

Issue #1137, opened 2013-04-16
Filter and ignore special characters in selection list


pkuczaj:
I have the following list in my chosen dropdown:
Rock
Rap
R&B
Jazz
Pop

When I type "R" my dropdown reduces to:
Rock
Rap
R&B

I would like to continue to type RB and reduce to:
R&B

At the moment I get a "No Results Found", BUT if system would ignore the & and any special characters to simplify and make the user experience better.

pfiller:
Something like #2618 would be right up your alley. Going to close this, but feel free to weigh in there.

PR #2618, opened 2016-06-03
Allow for a transliteration callback

hanoii:
Update 2016-10-06: This PR, while still open, is not currently being considered for merge. I need this on several projects and I want to reference the built package so that it can be pulled on different instances, so I will try to release packages from time to time. Those can be found on https://github.com/hanoii/chosen/releases. The code difference on those releases is what's referenced on this PR.

I needed a way to transliterate the items and search of what's on the dropdown. This made me do #2523 which was rightfully not accepted. I would still like to push a change that would allow me to do that without maintaining a custom fork of the project. Ir order to do so, I provided a harmless patch that simply adds the option to provide a transliteration callback.

As the previous pull request, this also attempts to fix or help with #536 .

Please double-check that:
- [x] You've updated both the jQuery _and_ Prototype versions.


public/options.html
```diff
@@ -130,6 +130,13 @@
             <p>By default Chosen's search is case-insensitive. Setting this option to <code class="language-javascript">true</code> makes the search case-sensitive.</p>
           </td>
         </tr>
+        <tr>
+          <td>transliteration_callback</td>
+          <td><code class="language-javascript">function(value) { return value }</code></td>
+          <td>
+            <p>By default, chosen performs no transliteration. This allows you to provide a callback performing your own alterations to the provided string argument. Callback receives a <code class="language-javascript">string</code> and must return a <code class="language-javascript">string</code>.</p>
```
tjschuck: Please capitalize "Chosen".

tjschuck:
This could actually be named even more generally, no?  I.e. there's nothing specifically tying this to transliteration.  It's actually just an option for a function to arbitrarily modify the search text.

I can't think of an actually good name, but it seems like naming this something like `search_text_transformation` instead would cover even more cases than just transliteration.

/cc @satchmorun, since he's good at naming things.

hanoii:
yes, I actually thought about that as well when doing this. Somehow I favoured transliteration because I couldn't really think of any other "useful" transformation to do, but what about:

normalisation_callback

I'd think that transliteration is a form of normalisation.

koenpunt:
It is not really a callback, so I think naming it `value_transformer` would make more sense IMO.

hanoii:
Well, it is a function that'd be called by chosen, so I think theoretically is a [callback](https://en.wikipedia.org/wiki/Callback_(computer_programming)), and I like the immediate reasoning that something that named xxx_callback would be likely a function, but I won't push this too hard, so let me know the naming of your preference and I'll update the PR.

koenpunt:
A callback is generally called on completion, not in the middle of a transaction. But let's wait for @satchmorun's opinion.

satchmorun:
I agree with @tjschuck, though i'd be more active tense about it:

`transform_search_text`

`normalize_search_text` would also be a-ok by me. `transform` is a little more general, but `normalization` is really the only thing this is ever going to get used for.

And since I've been pinged for comment, I'll add that line 159 should perform the regex escaping _after_ the normalization step, so that the transformation function doesn't accidentally break the regexification of the string. So something like: `escapedSearchText = this.normalize_search_text(option.search_text).replace(...)`

stof:
Does the replacement of matches actually work fine when using transliteration ? I'm not sure about it

tjschuck:
@stof What do you mean?

@hanoii Check out @satchmorun's last bit of feedback as well about the function call order.

hanoii:
Ok, attached are the proposed changes, @stof I am using it and just now updated my project with this change and works very good.

I took the liberty to modify the docs stylesheet very slightly so that the look of the table with the `<pre>` looks  better, otherwise it was adding an extra padding.

hanoii:
@tjschuck done. I leave you to fix the styling yourself as you see fit.

stof:
@hanoii the issue is that you search the startPos based on the normalized version to perform the highlighting replacement, which might be broken in case the normalization callback changes the length of the string (for instance, if you replace ``ö`` by ``oe`` to do German normalization of accents, the length changes and so positions will be shifted)

hanoii:
Nice call, let me se if there's something to do about that. Any ideas?

stof:
@hanoii no, I don't have an idea to fix it

hanoii:
Wow, @stof issue really set my brain on fire, if that's a valid english expression, but I couldn't stop thinking about this all day.

See if my latest commit is clear, it's a bit of a tricky logic, it's also probably not very cheap on iterations but I figured it shouldn't be that bad, and I made it so it only happens when an actual normalization happens on the search text.

I'll try to explain it with an example. And I actually gave #1137 a try as well as it was likely trigger some hind of highlighting issue as well.

Let's use the word:

"Speciꜵlwordr&bspeciꜵl"

The normalization converts the character ꜵ to "ao" and removes &.

Let's say we search for "iaolw" which should match "iꜵlw"

So I have two consecutive FORs, the first one start from the beginning of the word, and by removing the first character of the word on each iteration and normalizing it, I test whether the search is in the pattern. Because we are already there, we have a match for sure, all I am trying to find is when it stops matching, that would make the "startpos" of the real search_text regardless of any normalization.

So, some pseudo coding in the format i, str => normalize(str), still match?

Searching for "iaolw" (searchText is normalized as well)

0, "Speciꜵlwordr&bspeciꜵl" => "Speciaolwordrbspeciaol", yes
1, "peciꜵlwordr&bspeciꜵl" => "peciaolwordrbspeciaol", yes
2, "eciꜵlwordr&bspeciꜵl" => "eciaolwordrbspeciaol", yes
3, "ciꜵlwordr&bspeciꜵl" => "ciaolwordrbspeciaol", yes
4, "iꜵlwordr&bspeciꜵl" => "iaolwordrbspeciaol", yes
5, "ꜵlwordr&bspeciꜵl" => "aolwordrbspeciaol", no

Here the first for breaks and we know the start of the highlighting should happen at pos 4 of the real word

Now I do another for with a similar approach to find the length of the highlighting, starting from the previous found position and increasing the length, as soon as it matches the searchText, it means we found our length

So, pseudo coding in the format i, length, str => normalize(str), match?

4,0, "" => "", no
4,1, "i" => "i", no
4,2, "iꜵ" => "iao", no
4,3, "iꜵl" => "iaol", no
4,4, "iꜵlw" => "iaolw", yes

We found our length, 4.

So highlighting goes like this Spec<em>**iꜵlw**</em>ordr&bspeciꜵl

I tried a lot of use cases and seems to work very well.

BTW, found a very small bug, completely unrelated to this issue will write an issue but it's very late so just noting it here. If "search_contains" is not set, looking for a letter r in something like "Central African Republic" will match it fine but it will highlight the first r in Central. You can actually reproduce it right now in http://harvesthq.github.io/chosen/.

coffee/lib/abstract-chosen.coffee
```diff
@@ -178,13 +179,29 @@ class AbstractChosen
...
-              startpos = option.search_text.search zregex
-              text = option.search_text.substr(0, startpos + searchText.length) + '</em>' + option.search_text.substr(startpos + searchText.length)
+              startpos = this.normalize_search_text(option.search_text).search zregex
+              poslength = this.normalize_search_text(searchText).length
+
+              # If any normalization changed the search_text, perform a better
+              # logic to highlight results
+              if this.normalize_search_text(option.search_text) != option.search_text
+                for i in [0..option.search_text.length - 1] by 1
+                  if new RegExp(this.normalize_search_text(searchText), 'i').test(this.normalize_search_text(option.search_text.substr(i))) is false
```
stof: this is missing the escaping of special regex chars when building the regex

stof: btw, you should cache ``this.normalize_search_text(searchText)`` (and its escaping, and potentially even the regex itself) in a local variable instead of normalizing again for each iteration. This will be faster

coffee/lib/abstract-chosen.coffee
```diff
@@ -178,13 +179,29 @@ class AbstractChosen
...
+                    startpos = i - 1
+                    break
+
+                # this is not - 1 because i is used as length, not as index
+                for i in [0..option.search_text.length - startpos] by 1
```
stof: we should start the length at 1 IMO, not at 0

coffee/lib/abstract-chosen.coffee
stof: btw, we even already have the regex. It is the ``zregex`` variable

coffee/lib/abstract-chosen.coffee
```diff
@@ -178,13 +179,29 @@ class AbstractChosen
...
           if option.search_match
             if searchText.length
-              startpos = option.search_text.search zregex
-              text = option.search_text.substr(0, startpos + searchText.length) + '</em>' + option.search_text.substr(startpos + searchText.length)
+              startpos = this.normalize_search_text(option.search_text).search zregex
+              poslength = this.normalize_search_text(searchText).length
+
+              # If any normalization changed the search_text, perform a better
+              # logic to highlight results
+              if this.normalize_search_text(option.search_text) != option.search_text
```
stof: you should normalize the search_text only once (instead of doing it many times for each option) and the option search_text only once per option, to make things faster in case the normalization is a bit complex. It will be faster this way.

hanoii:
@stof done.

stof:
Looks good to me (and btw, I have an issue opened in my work project which would be a 5 min fix if this PR gets merged, so thank you)

tjschuck:
@hanoii Can you squash this down to one commit with a good message?  [Instructions here](http://blog.steveklabnik.com/posts/2012-11-08-how-to-squash-commits-in-a-github-pull-request).

Thanks!

hanoii:
@tjschuck done

tjschuck:
@pfiller Want to give this a last look?  We've got enough in the current release hopper as well that it's probably worth cutting a release shortly after this one goes in.

public/options.html
```diff
@@ -130,6 +130,15 @@
             <p>By default Chosen's search is case-insensitive. Setting this option to <code class="language-javascript">true</code> makes the search case-sensitive.</p>
           </td>
         </tr>
+        <tr>
+          <td>normalize_search_text</td>
+          <td><pre><code class="language-javascript">function(search_text) {
+  return search_text
+}</code></pre></td>
+          <td>
+            <p>By default, Chosen performs no text normalization. This allows you to provide a callback performing your own alterations to the provided search_text argument, i.e. for transliteration. Callback receives a <code class="language-javascript">string</code> and must return a <code class="language-javascript">string</code>.</p>
```
koenpunt: `search_text` should be in a `<code>` tag, I think

hanoii:
@koenpunt done and squashed.

koenpunt:
Even though I see the value in this, I still think there's too much logic involved for a feature that only certain people will use. 
Would it be possible to extract the highlight method as well? Then this logic has to be in the Chosen core

hanoii:
@koenpunt I don't honestly think there's that much of a logic, and it only kicks in when there is some normalization going on. I am really pushing this one to be accepted (please?), a lot of work has been put into it not to make it any way disruptive to the previous logic, and although the value might look like little, quite a few issues would be sorted and this expands the use of the library considerably for non-English speaking countries. And I was even planning on submitting some usage examples later on.

Although I appreciate this adds yet one more thing to maintain in the future once it gets in, not doing so would be a pity, is a nice feature and not too difficult to keep it in shape. I really don't see the harm in allowing such a feature in at this point and isn't this the core principle of open source contributions?

Going with your path, and looking for a way to have this also outside of chosen it's not easy. It uses a lot of the already used regex which those, or something in like, would have to be passed on to this "highlighting callback" you are suggesting, I think that's making this feature or its use even more complicated that it should be.

Anyway, I just think we reached a good simple working solution that would fit even more users to use this awesome library.

koenpunt:
What I mean is that the `winnow_results` function is already complex, and this adds more complexity to it, so if extract the highlighting from `winnow_results` to a `highlight_result` function, that could make it more readable. Agreed?

hanoii:
@koenpunt oh, I thought you meant outside of chosen. What about now?

hanoii:
Just a small bump to see if there's anything required/wanted for this to be merged? Thanks a lot!

hanoii:
Hi All! Just wondering if there's anything in particular holding this one back that I could help with.

koenpunt:
A little while back I talked with @pfiller about this PR, and we both agreed that we're not really happy with the complexity involved with the proposed changes. But we still have to setup a new meeting to discuss further development of Chosen, which preferably is more modular.

hanoii:
hmm, so that means this one is not going to be merged? Bummer! Do you have any suggestion in order to accept something that allows doing thing, I guess this is related to the modular talk, but if I can push something simpler..

What about simply providing an alternative winnow_results callback that you can work on your own? It could be feed with a context argument that's filled with necessary/useful information, like regex and such, or it could simply have the (this) sent as an argument..  something in the like?

hanoii:
So.. any comments on accepting maybe the possibility of providing a custom winnow_results callback? Almost no logic is needed for this, just new option.

koenpunt:
I think that how the `winnow_results` method is in your code is fine, as in; having the `highlight_option` separated from `winnow_results`. And if I'm correct, the only overridable method should then be `highlight_option`. Although making this method an option feels a bit off.. 
Any thoughts on this?

hanoii:
@koenpunt I am not sure that's quite right.. on my PR `winnow_results` also relies on `normalize_search_text` which is also proposed, and having to pass two different options to support such a functionality (`normalize_search_text` and `highlight_option`) does feel a bit off. 

However, removing `normalize_search_text` altogether and passing your own `winnow_result`, which could be slightly renamed to `process_result` or the like, would at least allow something like this (and possible other options) almost at no cost on the chosen side.

I still think the logic behind this PR is pretty good, and not that complicated, but I understand if you are reluctant to include something extra to maintain. Still, I really, really would like to push something that would allow me (and others) to do something like this without having to maintain a fork and keep it up to date upstream.

hanoii:
Namely, something like #2666

hanoii:
So, just following what @koenpunt mentioned on #2666, I see why something like that can be a really a one off, and just for me, so I closed that one.

Warning: very very small rant ahead.

On the other hand, this particular PR is not that heavy in terms on logic, and it does help with other issues mentioned here. I am not sure I am in the same page in rejecting this PR (yes, I am a bit biased) but trying to keep it objective, the PR doesn't introduce that much of heavy logic, and it only kicks in when this new feature (normalisation) is actually being used. This means that it doesn't affect or change the way people are using chosen right now, it only allows for a bit of further customisation, so it feels that's being rejected for "not liking it" than anything else.

That being said, I have a huge appreciation with any open source tool out there and I know you put hundreds if not thousands of hours into something like this, so thanks again.

So if there's really no further consideration for something like this, feel free to close this PR and I'll attempt to keep an up to date fork with matching releases to yours.

Thanks!

hanoii:
For what it's worth, I just updated the PR to the recent changes.

alejoar:
Just want to say thank you to @hanoii for keeping it up. Your release works great and we just switched to it in our production server. Cheers!

hanoii:
Not sure why the checks failed, but just updated to 1.7.0, rebased and released a new one with this included.

mdelorimier:
Hi,

Is there a directly installable version of you PR without compiling coffee script?

Regards :)

hanoii:
@mdelorimier as explained on the summary, I keep releases on https://github.com/hanoii/chosen/releases. I am not keeping up with chosen so quickly though. Will try to release a new one.

hanoii:
I spent quite a bit of time on this issue again. I kind of refactored everything into the newest code from scratch as rebasing was getting harder. Also somewhere in this PR it was requested I think by @koenpunt that the highlight logic should be moved to a separate function. I reverted that for the time being to make it easier to keep modifications up to date with master.

I also added an example to the main index page along the option documentation.

I still thing it's a nice addition to a great library. @tjschuck any reconsideration for something like this?

hanoii:
@mdelorimier a new release is up

tjschuck:
@harvesthq/chosen-developers Any of y'all have any followup thoughts on this?  Looks like @stof was on board, but then I paused for a final +1 from @pfiller.  We all know _that's_ not happening 😁 

@adunkman @satchmorun Any thoughts?  Particularly as this pertains to fixing #536, which seems like a pretty valid use case (with the caveat that I'm viewing #536 from the lens of a pretty-much ASCII-only language, so I don't know how real the problem of searching with `a` for `á` really is in practice).

koenpunt:
Well I’d say we’re missing tests. Also, if possible, a bit more code separation would do no harm; the `winnow_results` function is getting pretty big..

hanoii:
Tests, I wouldn't mind giving them a shot, probably with some assistance if you are willing, but I'd rather do that if/when this PR is accepted for merging.

On the same line, as mentioned above, I wouldn't mind separating `winnow_results`, but also rather do that either on a second issue after this gets in, because it really simplifies rebasing while chosen moves on its own.

Also if you checkout this branch and look into the docsupport you'll see a nice example of this working.

quentindemetz:
I would love to see this merged as well. French has a lot of accents and it's really useful to be able to search without typing them all in. @tjschuck is there anything I can do to help move this along? It's frustrating to see @hanoii 's solid work just sitting idle. 
Merci :fr:

garstud:
Hello, I have the same transliteration problem with **ChosenJS in version 1.6.2**.
I tried to integrate the methods **stringEscaped** and **winnow_results** from the Hanoi patch, replacing the original methods **winnow_results** and **get_search_regex**!
It works correctly and I haven't, for the moment, lifted any side effects that would cause bugs in ChosenJS.
Yet I am wary of using a Patch from 6 years ago.
- Do you know if the Hanoi patch causes incompatibilities in ChosenJs v1.6+?
- is there a reason why the transliteration problem has not been fixed in ChosenJS?

Thanks for your feedback.

stof:
> * is there a reason why the transliteration problem has not been fixed in ChosenJS?

The reason is given in the first lines of the readme: https://github.com/harvesthq/chosen#readme

garstud:
Thank you very much for your answer.
I'm not sure I understood. If it is related to the OpenSource contribution system, then I fully agree and the objective of my comment is precisely to advance the subject **through my feedback** and **oriented questions** to federate feedback from other people.
If the problem of transliteration support in ChosenJS is linked to a **lack of suitable PR**, is there a subject around the Hanoi patch that blocks its integration into a PR?
(With pleasure to bring my contribution on this subject in this beautiful project, if I can)

koenpunt:
There's no (active) development on this project; it's in a stable state, and no new features are currently accepted, so also this PR won't be merged any time soon (if ever). The team at Harvest has to decide on the future of Chosen. 
That said; if patching your version of Chosen works for you then I'd just continue doing that. I've actually also been using a modified copy for over 5 years in one of my projects.

garstud:
Ok, thank you for your explanations and your feedback. In this case, I will surely do like you and maintain a fork of ChosenJS for my specific project.

PR #2644, opened 2016-07-08
Disable Chosen on iPad, too.

thomasjfox:
Disable Chosen on the iPad, too, since f.e. scrolling on large lists doesn't work.

### DIFF
--- a/coffee/lib/abstract-chosen.coffee
+++ b/coffee/lib/abstract-chosen.coffee
@@ -286,7 +286,7 @@ class AbstractChosen
   @browser_is_supported: ->
     if "Microsoft Internet Explorer" is window.navigator.appName
       return document.documentMode >= 8
-    if /iP(od|hone)/i.test(window.navigator.userAgent) or
+    if /iP(ad|od|hone)/i.test(window.navigator.userAgent) or
        /IEMobile/i.test(window.navigator.userAgent) or
        /Windows Phone/i.test(window.navigator.userAgent) or
        /BlackBerry/i.test(window.navigator.userAgent) or


tjschuck:
@thomasjfox Sorry, what do you mean by "f.e. scrolling"?

thomasjfox:
On Friday, 8. July 2016 08:45:11 T.J. Schuck wrote:
> @thomasjfox Sorry, what do you mean by "f.e. scrolling"?

I had a multi-select list with a lot of entries. The iPad showed
the first XX entries and I couldn't scroll inside the list.

When I started typing some names, the entries showed up though.
So the data was loaded, I just couldn't scroll.

Hope that clears it up.

Cheers,
Thomas

tjschuck:
What version of iOS and what version of Chosen?

Can you confirm that the [Chosen demo page](http://harvesthq.github.io/chosen/) also shows the issue?

thomasjfox:
the device was running iOS 9.3.2 and Chosen plugin 1.5.1
together with jQuery 1.12.3.

I've run a git diff between 1.5.1 and 1.6.1 and the changes
looked minor to me.

I'll check the demo site on Monday.

Have a nice weekend,
Thomas

thomasjfox:
On Friday, 8. July 2016 09:01:52 T.J. Schuck wrote:
> Can you confirm that the [Chosen demo
> page](http://harvesthq.github.io/chosen/) also shows the issue?

yes, the demo site shows the same issue.

- you can't scroll the country list
- you can't remove already added entries (multi select)

Thomas

bensmoif:
Hey, we had the same issue.

I have a fix for this bug that i wrote for our library of Chosen (basically enhancing what is going on in the touchmove event), but I'm terribly inept at CoffeeScript so I haven't figured out how to submit a pullrequest with my fix.

Anyway, here's my file, works great on ipad.
[touchmovefix-chosen.jquery-1.5.1.js.txt](https://github.com/harvesthq/chosen/files/366800/touchmovefix-chosen.jquery-1.5.1.js.txt)

koenpunt:
I think disabling because the scroll isn't working isn't the way to go. We should investigate why the scroll isn't working, and eventually fix it.

PR #2346, opened 2015-06-16
Support for home and end key

ChristophKaser:
Allows pressing the home key to jump to the first entry of the selection, and pressing the end key to jump to the last.

PR #2650, opened 2016-07-20
Adding case keycode 36 for enabling POS1 selection for win users

mitchobrian:
It's just a little feature for windows users who are normally using html dropdown boxes with the pos1 key on their win keyboard to jump directly to the first item. For chosen I simulated it by deselecting the current item and scrolling to the top of the dropdown box.


koenpunt:
Thanks! Although just supporting POS1 (or Home) isn't enough; we would need "End", "Page up", and "Page down" too. Which I now found is already proposed in #2346, so closing in favor of that one.

PR #2666, opened 2016-08-10
process_results callback - PoC/WIP

hanoii:
### Summary

An alternative to #2618 that simply allows for a custom process results callback to be set.

If something like this is really going to be considered, I'll improve/work on the documentation.

Please double-check that:

  - [ ] If necessary, you've updated [the documentation](https://github.com/harvesthq/chosen/blob/master/public/options.html).

### References

- #2618
- #536

### DIFF
--- a/coffee/lib/abstract-chosen.coffee
+++ b/coffee/lib/abstract-chosen.coffee
@@ -34,6 +34,7 @@ class AbstractChosen
     @include_group_label_in_selected = @options.include_group_label_in_selected || false
     @max_shown_results = @options.max_shown_results || Number.POSITIVE_INFINITY
     @case_sensitive_search = @options.case_sensitive_search || false
+    @process_results = @options.process_results || false
 
   set_default_text: ->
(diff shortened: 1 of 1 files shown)


koenpunt:
I appreciate the eagerness of making Chosen better, but I think this is not very useful –except for you– because this requires a full reimplementation of the function, with option and option-group handling etc..
If only we make the Chosen object available, you can monkey-patch any of the methods, something like I started here: https://github.com/harvesthq/chosen/pull/1647

hanoii:
Well, for me and quite a few that commented on the parent issue, as well as other references to that issue as well. Although yes, I know, a handful. It was nice to see @davereid, a fellow Drupaler, commenting on #1647. I haven't really followed what's going on there but I must admin I haven't look properly, will do though.

Well, I guess it's really the end of the road for something like this to be accepted. Will just comment one more thing on the parent issue and we can call it/close it if there's really no more work around it to be considered.

Issue #2300, opened 2015-04-01
Search order and  search weight 


svenkatreddy:
First, thanks for the great library.

When i search for "india" in country list, the result order is a follows

British Indian Ocean Territory
India

Event though the second match is 100% it displays at 2nd position.

Is there some options to enable this feature or is it never implemented ?
Is there any prominence for the weight of the search string?

geoffroy-noel-ddh:
Hi, I would also be interested in this feature. Basically the ability to dynamically reorder the the suggestions based on the query and the matches. In this case if a match starts with the query it will be at the top. But if the user clears the query the order in the full list would be restored to its original state.

At least an API-friendly way to hook a dynamic post-match sorting function would be good.

PR #2668, opened 2016-08-16
Fix Search Only Working After 2nd Word

delprompt:
- Treating the result string to remove unnecessary spaces that may break the regexp search (Trim function used)

Probably related to : https://github.com/harvesthq/chosen/issues/2300


pfiller:
Thanks @delprompt. 

This is a tricky one, but I think we're going to close it. If trimming happens anywhere, it should probably happen at the parse level or when searching. Furthermore, there are some people who actually rely on the spaces to be there and this will break things for them.

PR #2669, opened 2016-08-17
Added styles to chosen.scss to fix #1410 Using a chosen-element alway…

### LINKED ISSUE #1410: Using a chosen-element always fires a horizontal page scroll to the left

rashmirao05:
on clicking the chosen element always fires a horizontal shift. To fix this used the following css. It fixes the issue in IE 11 and firefox. Please refer issue #1410 

opacity is not supported by all browsers, there will by a 1px shadow-line beneath the chosen-select. We tried it with Firefox 47 and IE11, both work fine.

We disable the left:-9999 so the focus() event is to going to scroll anywhere. display: none; would disable tabbing-support, so we do not hide the element, but make it invisible by height:0 and so that the 1px shadow disappears we use opacity: 0 but this is not supported by all browsers so they have to deal with 1px shadow…

For me, this solved the issue, but i did not test it with mobile devices.

### DIFF
--- a/sass/chosen.scss
+++ b/sass/chosen.scss
@@ -18,7 +18,10 @@ $chosen-sprite-retina: image-url('chosen-sprite@2x.png') !default;
   .chosen-drop {
     position: absolute;
     top: 100%;
-    left: -9999px;
+    left:0;
+ +  opacity:0;
+ +  height:0;
+ +  overflow: hidden;
     z-index: 1010;
     width: 100%;
     border: 1px solid #aaa;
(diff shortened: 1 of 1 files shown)


koenpunt:
I think using `clip: rect(0, 0, 0, 0)` could hide it completely without a shadow when opacity is not supported

rashmirao05:
This solution doesnt work im google chrome

koenpunt:
More info on hiding content: http://a11yproject.com/posts/how-to-hide-content/

pfiller:
I'm going to close this. I think we should try to investigate a solution along the lines of what @koenpunt is suggesting.

Thanks anyways, @rashmirao05.

PR #2574, opened 2016-03-16
Custom data-attributes for chosen generated options.

yhortuk:
This adds a possibility to pass custom data-attributes to generated chosen for further using in styling component etc.

PR #2679, opened 2016-08-31
Add option value to chosen's option element

liorch88:
This adds each of the option values to their chosen option element, as a "data-value" attribute. It also does the same for the selected option (or options) element.
It can help target specific chosen option elements and to apply css/js on them.
For example, to highlight a special option in the results, we could do something like:
```css
li[data-value=special] {
  color:red;
}
```
(without having to relay on the array index, which is subject to changes..)


koenpunt:
Thanks! Although a single data attribute is too specific. We plan on passing on all the options attributes to Chosen's options. Related #2574

PR #2687, opened 2016-09-12
add hide_results_on_select option

### LINKED ISSUE #1546: Stay open on multi select
Is there a way for the Chosen menu to stay open after I've selected an item? I want to be able to close it only by clicking outside the widget area.

koenpunt:
hide_results_on_select will default to true, to keep it backwards compatible.

the change is quite trivial, but the demand for it is high, so we might consider adding this?

cc: @pfiller @tjschuck @stof 

fixes #1546

### DIFF
--- a/coffee/chosen.jquery.coffee
+++ b/coffee/chosen.jquery.coffee
@@ -360,7 +360,7 @@ class Chosen extends AbstractChosen
       else
         this.single_set_selected_text(this.choice_label(item))
 
-      this.results_hide() unless (evt.metaKey or evt.ctrlKey) and @is_multiple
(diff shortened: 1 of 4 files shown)


tjschuck:
Seems okay to me 👍 

Up to you if you want a better 👍 than mine 😜

koenpunt:
I've drafted a release as 1.7.0 (minor b/c new feature). Are you aware of other fixes that we could include in this release? Or should we just release it?

boynet:
+1 for release it

koenpunt:
1.7.0 has just been released

deterset:
I like hide_results_on_select option but it always scrolls back to the top and I always have to find my place near the bottom or middle again. Is there any way to keep where you are so you can easily pick all the options you need ?

PR #2693, opened 2016-09-17
Support require and module.exports

saeedseyfi:
- Use `require('jquery')` instead of `window.jQuery` if `require` is available
- Pass `Chosen` to `module.exports` if it's possible
- Add `public/chosen.css`, `chosen.jquery.js` and `chosen.proto.js` + their minified versions to git 

Now developers can add the project as a `npm` package and easily require the js files (no need to install ruby, coffeescript and ... to use the project, just a require is enough)


saeedseyfi:
WIP

saeedseyfi:
Ready to review

pfiller:
Thanks @saeedseyfi.

I don't think this is able to merged as written, unfortunately. We've made an intentional effort to keep build files out of the repo. It opens us up to all kinds of confusing pull request issues and things. There's some weird tabbing issues going on as well. If you want to re-submit without all the noise, we'll review the actual change.

PR #2696, opened 2016-09-20
remove duplicate background property

koenpunt:
@mlettini this should be fine right?

### DIFF
--- a/sass/chosen.scss
+++ b/sass/chosen.scss
@@ -122,7 +122,6 @@ $chosen-sprite-retina: url('chosen-sprite@2x.png') !default;
       height: auto;
       outline: 0;
       border: 1px solid #aaa;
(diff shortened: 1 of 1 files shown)


pfiller:
@mlettini is on holiday at the moment so it'll be awhile before he weighs in.

I'm 👍 on this change. Chrome seems to be choosing the second rule (without the background color specified) so the line isn't even doing what we expect. I think this is an improvement.

Thanks, @koenpunt!

mlettini:
+1

koenpunt:
@mlettini valid point. Do we need the white background behind the image?

mlettini:
I don't think so, and again, the rule without the white background is the one that browsers have been using.

koenpunt:
Updated.

mlettini:
+1!

PR #2697, opened 2016-09-21
Input events

marcandre:
While not widely known, the `input` event is triggered before all `change` events for inputs and selects (see [the spec here](https://www.w3.org/TR/html5/forms.html#send-select-update-notifications))

This PR fixes this. Also adds a blurb on how to run tests...

I haven't changed the prototype version because it uses an old library that does not support the `input` event. Last commit was 4 years ago, looks like abandonware.

FWIW, `Parsley` relies on the `input` event, so the current code is [causing bad interactions](https://github.com/guillaumepotier/Parsley.js/issues/1134).

Thanks

### DIFF
--- a/coffee/chosen.jquery.coffee
+++ b/coffee/chosen.jquery.coffee
@@ -324,7 +324,7 @@ class Chosen extends AbstractChosen
     this.single_set_selected_text()
     this.show_search_field_default()
     this.results_reset_cleanup()
-    @form_field_jq.trigger "change"
+    this.trigger_form_field_change()
     this.results_hide() if @active_field
 
   results_reset_cleanup: ->
@@ -362,7 +362,7 @@ class Chosen extends AbstractChosen
       this.results_hide() unless @is_multiple && (!@hide_results_on_select || (evt.metaKey or evt.ctrlKey))
       this.show_search_field_default()
 
-      @form_field_jq.trigger "change", {'selected': @form_field.options[item.options_index].value} if @is_multiple || @form_field.selectedIndex != @current_selectedIndex
+      this.trigger_form_field_change selected: @form_field.options[item.options_index].value  if @is_multiple || @form_field.selectedIndex != @current_selectedIndex
       @current_selectedIndex = @form_field.selectedIndex
 
       evt.preventDefault()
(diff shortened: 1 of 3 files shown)


marcandre:
PS: If you really want, it's easy to update the library, but then it has to be stated that only the one you specify has to be used...

coffee/chosen.jquery.coffee
```diff
@@ -510,3 +510,7 @@ class Chosen extends AbstractChosen
         w = f_width - 10
 
       @search_field.css({'width': w + 'px'})
+
+  form_field_has_changed: (extra) ->
```
koenpunt: Also I'm not sure about the name of the method. More leaning to `trigger_form_field_change`, but maybe some of @harvesthq/chosen-developers can weigh in on this matter?

contributing.md
```diff
@@ -188,3 +188,13 @@ Once you're configured, building the JavaScript from the command line is easy:
     grunt watch                # watch coffee/ for changes and build Chosen
 
 If you're interested, you can find the task in [Gruntfile.coffee](https://github.com/harvesthq/chosen/blob/master/Gruntfile.coffee).
+
+#### Running tests
+
+To install all development dependencies, in the project's root directory, run
```
marcandre: It is indeed. As an experienced dev, the *only* thing I'm interested in is how do I get to run the tests. For this reason, I thought a separate (and independant) section was better. I probably chose the wrong header size. Let me tweak it and split this into another PR.

coffee/chosen.jquery.coffee
marcandre: `trigger_form_field_change` was my first idea :smile:, I thought it might be best to be a bit less specific in the name, but I'm fine with that name too.
You don't use `_myPrivateMethod` for private/protected methods, right? It's not really a public method.

marcandre:
OK, PR updated, with factorization done for prototype's version too.

koenpunt:
Looks good, thanks! 
@pfiller @stof @tjschuck second opinions?

koenpunt:
@marcandre would it be easily possible to feature-test the prototype version to see if it has support for `input`?

marcandre:
The only way would be to know is to try to call `simulate('input')` and rescue an exception...

The lib isn't all that useful anyways... Want a PR to get rid of it?

marcandre:
PR created, getting rid of `simulate` altogether.

tjschuck:
@koenpunt This looks good to me — if you're 👍 , I'm 👍

tjschuck:
@marcandre Do you mind squashing this down to one commit?  [Instructions if needed](http://blog.steveklabnik.com/posts/2012-11-08-how-to-squash-commits-in-a-github-pull-request).

marcandre:
Closing in favor of #2704

PR #2700, opened 2016-09-25
Contributing: add grunt command for tests [ci-skip]

marcandre:
A small edit to the 'grunt' section of the `Contributing` to list the important `grunt test` command.

As an experienced dev, the **only thing** I really need to know when wanting to write a PR is how to run the tests, and that was missing.

### DIFF
--- a/contributing.md
+++ b/contributing.md
@@ -175,7 +175,7 @@ license your work under the [MIT License](http://en.wikipedia.org/wiki/MIT_Licen
 4. Don't manually update the version number in `package.json`. This is done using a Grunt task on deployment.
 
 <a name="grunt"></a>
-#### Grunt tasks: Running Tests and using CoffeeScript
+#### Grunt tasks: Running Tests and building Chosen
 
 
 To install all development dependencies, in the project's root directory, run
--- a/contributing.md
+++ b/contributing.md
@@ -186,7 +186,7 @@ Once you're configured, `grunt` tasks are available:
 
     grunt test                 # run the tests in spec/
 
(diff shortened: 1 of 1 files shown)


contributing.md
```diff
@@ -175,16 +175,19 @@ license your work under the [MIT License](http://en.wikipedia.org/wiki/MIT_Licen
...
 
 
 To install all development dependencies, in the project's root directory, run
 
     npm install
 
-Once you're configured, building the JavaScript from the command line is easy:
+Once you're configured, `grunt` tasks are available:
+
+    grunt test                 # translate CoffeeScript and run the tests in spec/
```
koenpunt: Not sure about "translate". We speak of "build" later on, but maybe we should all replace that for "compile". 

> CoffeeScript is a little language that **compiles** into JavaScript.
> http://coffeescript.org

tjschuck: I would stick with "build Chosen and run the tests in spec/", and move this line to the second one _after_ `grunt build`, which implicitly specifies what "build Chosen" means.

marcandre: How about "run the tests in `spec/` from CoffeeScript source".

I wouldn't talk about 'build' since `grunt build` does a bunch of additional stuff.

koenpunt: I'm now thinking, is the addition "from CoffeeScript source" even necessary?

marcandre: Probably not. Removed.

koenpunt:
@tjschuck got anything to add to this?

Issue #2689, opened 2016-09-16
Activated Chosen instance can't be programmatically opened


koenpunt:
When a Chosen instance is activated using the `chosen:activate` event, a subsequent `chosen:open` event doesn't open the select anymore.

PR #2701, opened 2016-09-25
Allow activated instance to be programatically opened

IanSimpson:
### Summary

Fix for bug #2689 

Currently the `chosen:open` trigger piggybacks off the `container_mousedown` function. There was a check in here to prevent trying to open an instance two times if there were two clicks (by checking if the instance is already active), however this created this issue.

This check was changed to check if the instance was _showing_ instead, which seems to work.


koenpunt:
Thanks! Looks fine to me. @pfiller or @stof got anything to say about this?

Issue #2521, opened 2016-02-03
Inconsistent delete behavior


ramdog:
There is inconsistent behavior with respect to deleting a search.  Steps to reproduce:
### Flow A:
1. Search for something (non matched query)
2. Delete to the beginning (command + backspace)
3. Note the message the remains about an unmatched query
### Flow B:
1. Search for something (non matched query) - same as above^^^
2. Select all (command + A)
3. Delete (backspace)
4. Note it clears the search, which is different than the behavior as noted above

I would expect it to always work as Flow B.  Flow A seems like a bug to me.

Screencast taken directly from [demo page](https://harvesthq.github.io/chosen/):
![ulxzkbt7dc](https://cloud.githubusercontent.com/assets/5456178/12799112/fb85b072-ca81-11e5-9dee-5fdada0d8d40.gif)

tjschuck:
Thanks for the detailed report, @ramdog !

However, I can't reproduce :grimacing:  What browser, browser version, OS, and OS version are you using?

ramdog:
@tjschuck Thanks for the quick reply!  Good point, should have noted:

1.  Chrome Version `48.0.2564.97 (64-bit)`
2.  Mac OS X Yosemite Version `10.10.5 (14F1509)

tjschuck:
Hm, on the same version of Chrome, I'm not seeing the issue.  If I press cmd-backspace, as soon as I release command, it clears as I'd expect.

I'm on OS X El Cap, though, so perhaps that's the problem, though I doubt it would be...

ramdog:
Hmm, I just borrowed my cubemate's machine which is on El Cap (not sure on Chrome version) and I saw the same behavior (the buggy behavior).  

Also, when you say you can't reproduce, does that mean you only see the results in Flow A or Flow B?

tjschuck:
Same as your expectations -- Flow B seems correct; Flow A seems like a bug.  But for me, Flow A (cmd-backspace) also clears the search as expected.

ramdog:
Got it, interesting.  One thing I've noticed (without looking at the source yet) is that there seems to be some sort of debounce when updating the  `No results match "blahhhhhhhhhhhhhhh"` _but specifically on delete_.  In other words, if you:

1.  enter a long random (unmatched string), you'll note the "no results match" message, then
2.  hold down delete for a second or two to iteratively delete few characters in a row...

then I don't see the search update from:
 `No results match "blahhhhhhhhhhhhhhh"`

[...]

PR #2702, opened 2016-09-25
Make results_search run after keyup of Cmd key

IanSimpson:
### Summary

This is a fix for #2521 

Because of a [known behaviour](http://stackoverflow.com/questions/27380018/when-cmd-key-is-kept-pressed-keyup-is-not-triggered-for-any-other-key) on OSX, `keyup` events are not triggered for a key pressed while Cmd is being held down.

This lead to an issue when pushing Cmd-Delete on Mac to clear the search field. The `keyup` event handler ignores any pushes of (amongst others) modifier keys, including Ctrl, Alt, and Cmd.

This commit simply lets the search function continue to run after Cmd is released, so behaviour is closer to expected.

### DIFF
--- a/coffee/lib/abstract-chosen.coffee
+++ b/coffee/lib/abstract-chosen.coffee
@@ -254,7 +254,7 @@ class AbstractChosen
       when 27
         this.results_hide() if @results_showing
         return true
-      when 9, 38, 40, 16, 91, 17, 18
+      when 9, 38, 40, 16, 17, 18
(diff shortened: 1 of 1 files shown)


koenpunt:
Thanks! However I think the issue can also be solved by listening to the `input` event, instead of `keyup`. Using `input` has been proposed before as a fix for other issues, so we might look in to that, and maybe fix multiple at once.

Issue #2548, opened 2016-02-23
'search_contains' breaks search while escaping strings


ninboy:
Chosen now escapes correctly the quotes in the values (`'`), but there is a problem with the escape string while the option "search_contains" is active.

If you have an item with a quote the search will try to match against `&#x27;` and not `'`, making wrong results appear if you search for: `#`, `x`, `2`, `7` or `;`.

A small example is linked here: https://jsbin.com/luqusu/edit?html,output

Searching with `search_contains = false` makes the problem go away.

IanSimpson:
This issue is also present if an optgroup or option contains any html special character - >, <, &, etc. I've submitted a PR with a proposed fix.

Issue #1150, opened 2013-04-18
Find a way to safely match on options.text


pfiller:
When Chosen was built, searches matched on `options.text`, not on `options.html`. This created a possible [xss issue](https://github.com/harvesthq/chosen/issues/73) that needed to be addressed. You can read the thread for all the details, but the path chosen at the time was to match on `options.html`.

It is clear at this point that demand exists for matching on `options.text` (#581, #1119, #927, #905, #807) that we should find a way to do this safely (perhaps adding it as a configureable "use at your own risk" option).

Any PR that closes this issue must demonstrate clearly that it can fix the issues discovered in #73.

PR #2703, opened 2016-09-25
Make search match against text instead of html

IanSimpson:
### Summary

This is a fix for #2548.

With `search_contains` enabled, the search was matching the HTML. For example, if an optgroup had the name "This & that", a search for "m" would match the the m in &amp;.

This commit ensures searches are done against the raw text instead of the HTML value.

### DIFF
--- a/coffee/chosen.jquery.coffee
+++ b/coffee/chosen.jquery.coffee
@@ -403,7 +403,7 @@ class Chosen extends AbstractChosen
     @selected_item.addClass("chosen-single-with-deselect")
 
   get_search_text: ->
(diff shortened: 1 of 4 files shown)


koenpunt:
Thanks for your contributions! Over the years there has been quite a lot to do about matching html or text (eg. https://github.com/harvesthq/chosen/issues/1150 and https://github.com/harvesthq/chosen/pull/1638), so accepting a change like this isn't straightforward as you can imagine.

koenpunt:
And a bit more recent: #2254

PR #2704, opened 2016-09-25
Bye bye simulate

marcandre:
This PR removes the dependency on `simulate`, which is old, unmaintained and incomplete.

Replacing costs about 5 lines of code...

It also adds the `input` event to the prototype version.

Builds on #2697

### DIFF
--- a/coffee/chosen.proto.coffee
+++ b/coffee/chosen.proto.coffee
@@ -506,6 +506,7 @@ class @Chosen extends AbstractChosen
       @search_field.setStyle({'width': w + 'px'})
 
   trigger_form_field_change:  ->
+    triggerHtmlEvent @form_field, 'input'
     triggerHtmlEvent @form_field, 'change'
 
   triggerHtmlEvent = (element, eventType) ->
--- a/coffee/chosen.proto.coffee
+++ b/coffee/chosen.proto.coffee
@@ -315,7 +315,7 @@ class @Chosen extends AbstractChosen
     this.single_set_selected_text()
     this.show_search_field_default()
     this.results_reset_cleanup()
-    @form_field.simulate("change") if typeof Event.simulate is 'function'
+    this.trigger_form_field_change()
     this.results_hide() if @active_field
 
   results_reset_cleanup: ->
@@ -353,7 +353,7 @@ class @Chosen extends AbstractChosen
       this.results_hide() unless @is_multiple && (!@hide_results_on_select || (evt.metaKey or evt.ctrlKey))
       this.show_search_field_default()
 
-      @form_field.simulate("change") if typeof Event.simulate is 'function' && (@is_multiple || @form_field.selectedIndex != @current_selectedIndex)
+      this.trigger_form_field_change() if @is_multiple || @form_field.selectedIndex != @current_selectedIndex
       @current_selectedIndex = @form_field.selectedIndex
 
       evt.preventDefault()
@@ -381,7 +381,7 @@ class @Chosen extends AbstractChosen
       this.result_clear_highlight()
       this.winnow_results() if @results_showing
 
-      @form_field.simulate("change") if typeof Event.simulate is 'function'
+      this.trigger_form_field_change()
       this.search_field_scale()
       return true
     else
@@ -504,3 +504,6 @@ class @Chosen extends AbstractChosen
         w = f_width - 10
 
       @search_field.setStyle({'width': w + 'px'})
+
+  trigger_form_field_change:  ->
+    @form_field.simulate("change") if typeof Event.simulate is 'function'
--- a/coffee/chosen.proto.coffee
+++ b/coffee/chosen.proto.coffee
@@ -506,4 +506,15 @@ class @Chosen extends AbstractChosen
       @search_field.setStyle({'width': w + 'px'})
 
   trigger_form_field_change:  ->
-    @form_field.simulate("change") if typeof Event.simulate is 'function'
+    triggerHtmlEvent @form_field, 'change'
+
+  triggerHtmlEvent = (element, eventType) ->
+    if element.dispatchEvent  # Modern way:
+      try
+        evt = new Event(eventType, bubbles: true, cancelable: true)
+      catch
+        evt = document.createEvent('HTMLEvents')
+        evt.initEvent(eventType, true, true);
+      element.dispatchEvent(evt)
+    else # Old IE:
+      element.fireEvent("on#{eventType}", document.createEventObject());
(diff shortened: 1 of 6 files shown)


koenpunt:
👍 Looks good, will review in the morning.

coffee/chosen.proto.coffee
```diff
@@ -504,3 +504,14 @@ class @Chosen extends AbstractChosen
         w = f_width - 10
 
       @search_field.setStyle({'width': w + 'px'})
+
+  trigger_form_field_change:  ->
+    Chosen.trigger(@form_field, 'input')
+    Chosen.trigger(@form_field, 'change')
+
+  @trigger: (element, eventType) ->
```
koenpunt: Having `trigger` as public method shouldn't be necessary. But this probably has been done so the method is available in the tests as well?

coffee/chosen.proto.coffee
```diff
@@ -504,3 +504,14 @@ class @Chosen extends AbstractChosen
...
+    if (document.createEventObject) # Old IE:
```
koenpunt: `if` statements should be written without parens (I know, there are some conditions that do use parens, but they just slipped through the cracks I guess)

spec/proto/basic.spec.coffee
```diff
@@ -27,11 +27,11 @@ describe "Basic setup", ->
     expect($F(select)).toBe ""
 
     container = div.down(".chosen-container")
-    container.simulate("mousedown") # open the drop
+    Chosen.trigger(container, "mousedown") # open the drop
     expect(container.hasClassName("chosen-container-active")).toBe true
 
     #select an item
-    container.select(".active-result").last().simulate("mouseup")
+    Chosen.trigger(container.select(".active-result").last(), 'mouseup');
```
koenpunt: Nitpicky, but inconsistent quote usage. On multiple locations, but marking them all feels too extreme.

And yes I know it's hypocritical, because we use single and double quotes randomly around the library, but I still strive for some consistency.

coffee/chosen.proto.coffee
marcandre: Right. Should I call it `_trigger` instead?

spec/proto/basic.spec.coffee
```diff
@@ -27,11 +27,11 @@ describe "Basic setup", ->
     expect($F(select)).toBe ""
 
     container = div.down(".chosen-container")
-    container.simulate("mousedown") # open the drop
+    Chosen.trigger(container, "mousedown") # open the drop
```
stof: this is not a good idea, because ``Chosen.trigger`` is not a full-fledged implementation of triggering builtin events (it always relies on the Event class, while many DOM events are using subclasses with extra properties)

spec/proto/basic.spec.coffee
marcandre: :+1: should not have switched them

coffee/chosen.proto.coffee
```diff
@@ -504,3 +504,14 @@ class @Chosen extends AbstractChosen
...
+      element.fireEvent("on#{eventType}", document.createEventObject());
+    else # Modern way:
```
stof: you should detect support for the modern way instead (detecting ``dispatchEvent``), to use the modern way when multiple ways are available. ``fireEvent`` should be a fallback only.

Btw, you should also write the code using ``createEvent()`` and ``initEvent`` when ``dispatchEvent`` is available but not the Event constructor (i.e. the intermediate API)

marcandre:
Tweaked as per comments, and reorganized the commits to make more sense.

spec/proto/basic.spec.coffee
marcandre: So what is a better idea?

stof: Well, as Chosen does not need a full-fledged event triggering system (as it does not need to trigger mouse events), the simpler would probably be to keep using an external library for that in the testsuite

marcandre: I arrived at the same conclusion :smile:

marcandre:
Ok, hopefully third time's the charm.

No more `Chosen.trigger`, dependency on `simulate` is kept only for testing, and prototype version supports 5 year old Safari and al., while being forward compatible by not using `initEvent` and al. when possible (those are deprecated).

coffee/chosen.proto.coffee
```diff
@@ -504,3 +504,19 @@ class @Chosen extends AbstractChosen
...
+  triggerHtmlEvent = (element, eventType) ->
+    if element.dispatchEvent  # Modern way:
+      try
+        evt = new Event(eventType, bubbles: true, cancelable: true)
+      catch
+        evt = document.createEvent('HTMLEvents')
+        evt.initEvent(eventType, true, true);
+      element.dispatchEvent(evt)
+    else
+      document.createEventObject # Old IE:
```
stof: this line is useless

public/options.html
```diff
@@ -229,7 +229,7 @@
           <td>change</td>
           <td>
             <p>Chosen triggers the standard DOM event whenever a selection is made (it also sends a <code class="language-javascript">selected</code> or <code class="language-javascript">deselected</code> parameter that tells you which option was changed).</p>
-            <p><strong>Note:</strong> in order to use change in the Prototype version, you have to include the <a href="https://github.com/kangax/protolicious/blob/5b56fdafcd7d7662c9d648534225039b2e78e371/event.simulate.js">Event.simulate</a> class. The selected and deselected parameters are not available for Prototype.</p>
+            <p>The selected and deselected parameters are not available for Prototype.</p>
```
stof: ``<strong>Note:</strong>`` should be kept

marcandre: Fixed

coffee/chosen.proto.coffee
marcandre: Indeed. Fixed

koenpunt:
Thanks again, @marcandre 💯

marcandre:
Is it possible to make a release with this?

koenpunt:
I'm collecting as much small bugfixes as possible for a new release, but I expect to release 1.7.0 somewhere next week.

marcandre:
I would still very much appreciate a release...

koenpunt:
@marcandre I understand, and it has been more than a week now, but there are still quite some PR's that I'd like to see merged for the 1.7.0 release.
Truth is though that I can't make all decisions on my own, and not all maintainers are as active as I have been for the past month (which is fine, because open source), which causes that not everything goes as quick as I want it to go.
But if you like, reviews on the last 8 pull requests are more than welcome, which can eventually speed up the review process for the other maintainers.

koenpunt:
@marcandre took some time, but as there was no progress since a few months I decided to put out a release. So 1.7.0 is now available.

marcandre:
Thanks @koenpunt

Issue #2370, opened 2015-07-16
A way to get user typed text on "chosen:no_results"


VFiber:
If you define a handler for 'chosen:no_results' event its pretty tricky to get the text (what provided the 'no result') from user input.

Could you please add a function (or extra parameter to the callback) to make it possible to get the text from the input field?

kylehayes:
+1

adamnbowen:
Seconded.  It seems that [the current method](https://github.com/harvesthq/chosen/blob/910a12599c10a11b7005feedb1e53e4b4535100e/coffee/chosen.jquery.coffee#L413-L418) doesn't actually pass the terms through, though I was able to grab them in a super awkward way:

```javascript
$('.select--chosen').on('chosen:no_results', function(event, parameters) {
    console.log(parameters.chosen.search_results[0].textContent.match(/No results match "(.+)"/)[1])
  });
```

dzeju555:
Has anyone any idea how to solve that problem?

adamnbowen:
@dzeju555 which problem? The code I posted above makes it possible to grab the text the user typed, it's just a bit awkward.  Otherwise, a pull request needs to make changes to the library to give us an easier way to get at the data.

dzeju555:
I mean, the easier way to do that. I have to do a selectlist which allows user to select an option. But if there is no option user typed, I have to add it to a database dictionary. I though that it will be a good option that Chosen could sent typed text in form with another input fields and in my controller (I use MVC application) I would do the rest. What is important, I want typed text to be still in Chosen field, now it disappears when Chosen is closed and user is not sure what will be sent to the database. I will try to replace placeholder with typed text in jQuery when not found trigger is ran.

Modifikator:
```
var obs = $('select').chosen({width: '100%'}).data('chosen');
            obs.search_field.on('keyup', function(e){
                console.log($(this).val());
            });
```

niktrix:
+1

joao-bjsoftware:
Maybe this can works.
      
`
$('.select-chosen').on('chosen:no_results', function(event, element){
   console.log($(element.chosen.search_results).find('span').text());
});
`

koenpunt:
This is already possible:

```js
$('.chosen-select').on('chosen:no_results', function(event, data){
  console.log(data.chosen.get_search_text());
});
```

PR #2713, opened 2016-10-02
prevent bubbling of escape key

### LINKED ISSUE #1256: ESC/Escape key closes containing jQueryUI dialog when Chosen drop-down list is showing [jsFiddle provided]
When using a Chosen drop-down list inside of a jQueryUI dialog, an issue exists with the ESC/Escape key where if the user presses ESC while the drop-down list is showing (ie. results_showing==true) then instead of the drop-down list disappearing as expected, the entire containing dialog is canceled.

Please refer to [this jsFiddle](http://jsfiddle.net/LatencyMachine/KngjW/) for a demonstration of the issue and a workaround fix.

This appears to stem from the dialog looking for ESC on keydown whereas Chosen is looking for ESC on keyup.  

Hopefully there is a more elegant way to resolve this, ideally from within Chosen and not externally as shown in my demo.

### LINKED ISSUE #2653: issue when usin bootstrap modal esc keypress
i found a solution , hopes implement on future release; 
the situation is when using bootstrap modal, and shows chosen item, if press esc key, both modal and chosen close,,, expected only chosen hide, and modal stills i modifed 
i use stoppropagation.
the calls con keyup_checker and keydown_checker and put this lines:
[keydown_checker]
case 27:
          if (this.results_showing) {
            evt.stopPropagation();
            this.results_hide();
          }
          return true;

[keyup_checker]
case:27
if (this.results_showing) {
            evt.stopPropagation();
            this.results_hide();
          }
          return true;

koenpunt:
When the results are showing, pressing escape only hides the results, and then prevents bubbling.

Also introduced an extra method to retrieve the search input value, so that the `keydown_checker` can be moved to the abstract class. 
This method can also be used as an additional solution for #2370, e.g.:

```js
$('.chosen').chosen().on('chosen:no_results', function(event, data) {
  console.log(data.chosen.get_search_field_value());
});
```

fixes #1256 
fixes #2653

### DIFF
--- a/coffee/lib/abstract-chosen.coffee
+++ b/coffee/lib/abstract-chosen.coffee
@@ -271,21 +271,25 @@ class AbstractChosen
     this.search_field_scale()
 
     switch stroke
-      when 8
+      when 8 # backspace
         if @is_multiple and @backstroke_length < 1 and this.choices_count() > 0
           this.keydown_backstroke()
         else if not @pending_backstroke
           this.result_clear_highlight()
           this.results_search()
-      when 13
+        break
+      when 13 # enter
         evt.preventDefault()
         this.result_select(evt) if this.results_showing
-      when 27
+        break
+      when 27 # escape
         this.results_hide() if @results_showing
-        return true
-      when 9, 38, 40, 16, 91, 17, 18
+        break
+      when 9, 16, 17, 18, 38, 40, 91
         # don't do anything on these keys
-      else this.results_search()
+      else
+        this.results_search()
+        break
 
   clipboard_event_checker: (evt) ->
     setTimeout (=> this.results_search()), 50
--- a/coffee/lib/abstract-chosen.coffee
+++ b/coffee/lib/abstract-chosen.coffee
@@ -237,6 +237,35 @@ class AbstractChosen
     evt.preventDefault()
     this.results_show() unless @results_showing or @is_disabled
 
+  keydown_checker: (evt) ->
+    stroke = evt.which ? evt.keyCode
+    this.search_field_scale()
+
+    this.clear_backstroke() if stroke != 8 and @pending_backstroke
+
+    switch stroke
+      when 8 # backspace
+        @backstroke_length = this.get_search_field_value().length
+        break
+      when 9 # tab
+        this.result_select(evt) if @results_showing and not @is_multiple
+        @mouse_on_container = false
+        break
+      when 13 # enter
+        evt.preventDefault() if @results_showing
+        break
+      when 32 # space
+        evt.preventDefault() if @disable_search
+        break
+      when 38 # up arrow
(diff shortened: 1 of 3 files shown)


coffee/lib/abstract-chosen.coffee
```diff
@@ -237,26 +237,62 @@ class AbstractChosen
...
         evt.preventDefault()
         this.result_select(evt) if this.results_showing
-      when 27
+        break
+      when 27 # escape
         this.results_hide() if @results_showing
-        return true
-      when 9, 38, 40, 16, 91, 17, 18
+        break
+      when 9, 16, 17, 18, 38, 40, 91
```
tjschuck: My hero.

coffee/lib/abstract-chosen.coffee
```diff
@@ -237,26 +237,62 @@ class AbstractChosen
...
     switch stroke
-      when 8
+      when 8 # backspace
         if @is_multiple and @backstroke_length < 1 and this.choices_count() > 0
           this.keydown_backstroke()
         else if not @pending_backstroke
           this.result_clear_highlight()
           this.results_search()
-      when 13
+        break
```
tjschuck: You don't need to add all of these explicit `break`s here; CoffeeScript handles that.

[Docs](http://coffeescript.org/#switch):

> Switch statements in JavaScript are a bit awkward. You need to remember to break at the end of every case statement to avoid accidentally falling through to the default case. CoffeeScript prevents accidental fall-through...

Confirmed with the "Try CoffeeScript" dingus on http://coffeescript.org/ as well.

coffee/lib/abstract-chosen.coffee
```diff
@@ -237,26 +237,62 @@ class AbstractChosen
...
+  keydown_checker: (evt) ->
+    stroke = evt.which ? evt.keyCode
+    this.search_field_scale()
+
+    this.clear_backstroke() if stroke != 8 and @pending_backstroke
+
+    switch stroke
+      when 8 # backspace
+        @backstroke_length = this.get_search_field_value().length
+        break
```
tjschuck: Ditto on all of these `break`s — can be safely removed.

coffee/lib/abstract-chosen.coffee
koenpunt: I know, but if you add them, CoffeeScript returns a value, and I wanted to get rid of the `return true` below, so for consistency I added it everywhere.

tjschuck: Gotcha.  Fair enough.

tjschuck:
The biggest question I had that came out of reviewing this is WTF a `backstroke` is, but it looks like that name was added in the _initial commit_ 😮 by @pfiller 

👍 on this one, though!

PR #2714, opened 2016-10-02
add white-space: pre for search field scaling

### LINKED ISSUE #791: Search field doesn't scale with spaces
If you type in few spaces into search field it won't scale, because of this line (chosen / chosen / chosen.jquery.js:968):

    div.text(this.search_field.val());
It'll work with

    div.html(this.search_field.val().replace(/\s/g, '&nbsp;'));

koenpunt:
this prevents spaces to be collapsed

fixes #791


mlettini:
+1

PR #2715, opened 2016-10-02
add shortcut for deselect

### LINKED ISSUE #994: More shortcut keys to allow for keyboard only selection deselection
Great work here guys.

This request is to allow add a few more shortcut keys or allow a way that the keys may be customized. 

1) When you have a single deselect option allow the delete key to remove the selection when the drop had focus

case 46:
        if (this.allow_single_deselect && !this.results_showing && this.current_value!="") {
            this.results_reset(evt);
         }
         return true;

2) When drop has focus hitting return opens drop.

case 13:
evt.preventDefault();
          if (this.results_showing) {
              return this.result_select(evt);
          } else {
              this.results_show();
          }

3) Also seems to when using the arrow keys to open the drop when it has focus, I think that it feels natural to close it with the up arrow. So say when the search textbox has focus and the up arrow is pressed.

### LINKED ISSUE #2563: Allow Deselect on Single Selects using keyboard (esc key?)
In the case of a single select that allows deselection it would be a usability improvement to be able to do this using only the keyboard. Currently we need to use the mouse to click on the x. I propose to allow deselecting of a selected single select by hitting the esc key.

koenpunt:
pressing delete will deselect when `@allow_single_deselect` is `true`.

fixes #994 
fixes #2563

### DIFF
--- a/coffee/lib/abstract-chosen.coffee
+++ b/coffee/lib/abstract-chosen.coffee
@@ -270,6 +270,11 @@ class AbstractChosen
         evt.preventDefault()
         this.keydown_arrow()
         break
+      when 46 # delete
+        if @allow_single_deselect and not @results_showing
+          evt.preventDefault()
+          this.results_reset()
+        break
 
   keyup_checker: (evt) ->
     stroke = evt.which ? evt.keyCode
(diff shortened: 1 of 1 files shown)


tjschuck:
Just a note that #2713 (now +1'd, probably merged shortly) is going to break this diff and rebasing will be needed.

koenpunt:
This might be in need of some additional documentation

PR #2718, opened 2016-10-02
consider disabled fieldset for disabled state

### LINKED ISSUE #2389: Disabling fieldset doesn’t disable its descendants.
Form controls within a disabled fieldset should be disabled: https://www.w3.org/TR/html5/forms.html#attr-fieldset-disabled

Chosen does not seem to support this case: http://codepen.io/arkhi/full/rVPddZ/

koenpunt:
fixes #2389

### DIFF
--- a/coffee/chosen.jquery.coffee
+++ b/coffee/chosen.jquery.coffee
@@ -115,14 +115,8 @@ class Chosen extends AbstractChosen
     @form_field_jq.show()
 
   search_field_disabled: ->
-    @is_disabled = @form_field.disabled || (=>
-      element = @form_field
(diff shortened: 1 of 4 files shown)


koenpunt:

koenpunt:
I'm only thinking that I should move the loop to a separate function, as it is the same both for jQuery and prototype.

koenpunt:
I've now updated it to use jQuery/Prototype specific methods.

And if it wasn't for _good ol' IE_, the jQuery version could be as simple as:

```coffeescript
@is_disabled = @form_field_jq.is(':disabled')
```

Issue #2575, opened 2016-03-18
Allow a way to not have the width attribute set inline on the chosen element


stringbeans:
The width attribute is always set inline on the chosen element and this causes issues around styling and forcing styles to use !important.

It would be nice if there was a way to turn off the default behaviour of setting the width inline so that we can properly style the chosen element width with css

pfiller:
I think this is probably something we should take a look at, particularly since CSP and inline styles don't play nicely (#2423).

TadWohlrapp:
I had the same issue, fixed it this way:
```
$('.my_select').chosen({
  options
}).next('.chosen-container').css('width', '');
```
It simply looks for the .chosen-container following your select-element and removes its inline css.

PR #2723, opened 2016-10-15
Content-Security-Policy fixes

### LINKED ISSUE #2423: Inline styles and Content Security Policy
As of now, Chosen for JQuery does not work correctly when inline styles usage is forbidden by an appropriate Content-Security-Policy.
The problem is that, when it creates the "chosen" menu in set_up_html, it uses "style" param to specify width: https://github.com/harvesthq/chosen/blob/910a12599c10a11b7005feedb1e53e4b4535100e/coffee/chosen.jquery.coffee#L35
JQuery supports specifying the width directly (not by CSS properties): http://api.jquery.com/width/ ; this one should solve the problem.

Unfortunately, writing front-end JS compatible across different browsers (and testing it) is not my strong side, so I'll only create this issue, and will not supply any pull request with it.

koenpunt:
To make Chosen compatible with restrictive Content-Security-Policy's, I've updated the way we apply inline styles to elements.

The HTML used to build the select is now defined in AbstractChosen.

I also updated the documentation pages with a CSP header as an example.

fixes #2423 
related #2575

### DIFF
--- a/coffee/lib/abstract-chosen.coffee
+++ b/coffee/lib/abstract-chosen.coffee
@@ -333,6 +333,39 @@ class AbstractChosen
     tmp.appendChild(element)
     tmp.innerHTML
 
+  get_single_template: ->
+    """
+      <a class="chosen-single chosen-default">
+        <span>#{@default_text}</span>
+        <div><b></b></div>
+      </a>
+      <div class="chosen-drop">
+        <div class="chosen-search">
+          <input class="chosen-search-input" type="text" autocomplete="off" />
+        </div>
+        <ul class="chosen-results"></ul>
+      </div>
+    """
+
+  get_multi_template: ->
+    """
+      <ul class="chosen-choices">
+        <li class="search-field">
+          <input class="chosen-search-input" type="text" autocomplete="off" value="#{@default_text}" />
+        </li>
+      </ul>
+      <div class="chosen-drop">
+        <ul class="chosen-results"></ul>
+      </div>
+    """
+
+  get_no_results_template: (terms) ->
+    """
+      <li class="no-results">
+        #{@results_none_found} <span>#{terms}</span>
+      </li>
+    """
+
   # class methods and variables ============================================================
 
   @browser_is_supported: ->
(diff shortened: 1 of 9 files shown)


koenpunt:

public/index.proto.html
```diff
@@ -6,9 +6,25 @@
   <link rel="stylesheet" href="docsupport/style.css">
   <link rel="stylesheet" href="docsupport/prism.css">
   <link rel="stylesheet" href="chosen.css">
+
+  <meta http-equiv="Content-Security-Policy" content="default-src 'self'; script-src 'self' https://ajax.googleapis.com; style-src 'self' 'sha256-MzlZlLruPmzLm/yy8z6yCpmyo5QSe/SURdcVrrJx880='; img-src 'self' data:">
```
stof: using a nonce rather than a hash may make maintenance easier, by avoiding to recompute the hash when updating the demo CSS (which will be forgotten for sure)

coffee/lib/abstract-chosen.coffee
```diff
@@ -326,6 +326,39 @@ class AbstractChosen
...
+        <li class="search-field">
+          <input class="chosen-search-input" type="text" autocomplete="off" value="#{@default_text}" />
+        </li>
+      </ul>
+      <div class="chosen-drop">
+        <ul class="chosen-results"></ul>
+      </div>
+    """
+
+  get_no_results_template: (terms) ->
```
stof: this is not really returning a template. It is *rendering* a template

koenpunt: Yeah thought of that. Same applies for `get_multi`/`get_single`. Maybe a suffix of `_html` is better.

public/index.proto.html
koenpunt: Even though calculating the hash is easy, because the correct one is logged to the console when using an incorrect hash, the forgetting part is true. 

But then again, a nonce isn't safe if not regenerated on every request. Not that there's really a security risk at this page, I still think we shouldn't advertise the less secure option.

Maybe we can move the CSP additions to the build process of the documentation, and leave out the header in the HTML here.

Or validate the HTML in the build process, if possible.

koenpunt: Or better yet, we can just move the rules to `docsupport/style.css`

PR #2725, opened 2016-10-15
only prevent default on touchevents when results are not showing

### LINKED ISSUE #2289: Chosen on Ipad scroll problem single select en click problem multiple select
The recent fix:
#2119 Fix touch event propagation
seems to breaks other functionality of chosen on Ipad (tested in Ipad 3, latest version chosen).
Problems:
 - single select: scroll of dropdown list does not work any more
 - multiple select: removing selected items is impossible and after the first time a click in the select window does not show the drop down list
These problems can be reproduced using an Ipad on the standard chosen test site: http://harvesthq.github.io/chosen/  on the Standard select and Multiple select examples.
They probably have to do with the #2119 Fix as the functionality was ok before.

### LINKED ISSUE #2443: Does not work on Android tablets (e.g. Nexus 9)

koenpunt:
The fix introduced in #2119 prevented all further touch interaction with Chosen like scrolling or selecting options.

I don't know how that has been in there for so long...

I've now updated that to only `preventDefault` the events when the results are not showing.

fixes #2289
fixes #2443

### DIFF
--- a/coffee/chosen.jquery.coffee
+++ b/coffee/chosen.jquery.coffee
@@ -69,8 +69,8 @@ class Chosen extends AbstractChosen
     @form_field_jq.trigger("chosen:ready", {chosen: this})
 
   register_observers: ->
-    @container.bind 'touchstart.chosen', (evt) => this.container_mousedown(evt); evt.preventDefault()
-    @container.bind 'touchend.chosen', (evt) => this.container_mouseup(evt); evt.preventDefault()
+    @container.bind 'touchstart.chosen', (evt) => this.container_mousedown(evt); return
+    @container.bind 'touchend.chosen', (evt) => this.container_mouseup(evt); return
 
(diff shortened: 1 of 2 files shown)


koenpunt:

tjschuck:
LGTM 👍

jessarcher:
Any chance of releasing a package containing this fix sometime soon? It would be much appreciated.

koenpunt:
v1.7.0 has just been released

PR #2726, opened 2016-10-15
use a tag instead of abbr tag for search-choice-close

### LINKED ISSUE #495: Do not use <abbr> tag
Nevermind that using semantic tags as decoration elements isn't a good practice, if some styles globally set for abbr tag, they affects chosen's selection clearing cross. To be specific, Bootstrap 2.0 defines some, so using bootstrap and chosen together means resetting abbr styles specifically for chosen.

koenpunt:
fixes #495

### DIFF
--- a/coffee/chosen.proto.coffee
+++ b/coffee/chosen.proto.coffee
@@ -290,7 +290,7 @@ class @Chosen extends AbstractChosen
     if item.disabled
       choice.addClassName 'search-choice-disabled'
     else
-      close_link = new Element('button', { type: 'button', class: 'search-choice-close', rel: item.array_index })
+      close_link = new Element('button', { type: 'button', tabindex: -1, class: 'search-choice-close', rel: item.array_index })
       close_link.observe "click", (evt) => this.choice_destroy_link_click(evt)
       choice.insert close_link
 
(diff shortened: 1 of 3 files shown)


koenpunt:
@mlettini do you have anything to say about this?

mlettini:
* +1 to not using the `abbr` tag for this... forget why it was done in the first place!
* Now that we're using an `<a>`, I think we need to add more reset properties to `.search-choice-close` just to make sure that a page's base styles don't break it, something like: `{ margin:0; padding:0; border:0; }`, though potentially other properties too...
* The above is not a +1 to the JS which needs another dev to look at it 👌

coffee/chosen.proto.coffee
```diff
@@ -392,8 +392,9 @@ class @Chosen extends AbstractChosen
 
   single_deselect_control_build: ->
     return unless @allow_single_deselect
-    @selected_item.down("span").insert { after: "<abbr class=\"search-choice-close\"></abbr>" } unless @selected_item.down("abbr")
-    @selected_item.addClassName("chosen-single-with-deselect")
+    unless @selected_item.down('.search-choice-close')
+      @selected_item.down('span').insert { after: '<a class="search-choice-close"></a>' }
```
stof: as there is no href, this won't get a pointer cursor in a cross-browser way (and browsers not showing it are perfectly fine according to the spec)

mlettini: Lines 28-30 of `sass/chosen.scss` force a cursor to all links within `.chosen-container`, but I agree that no href is less semantic.

stof: proper semantic would probably be a ``<button>`` rather than a ``<a>`` btw.

koenpunt:
Updated per @stof's comment about using a button. Removed some duplicate CSS in the process.

PR #2727, opened 2016-10-15
clip results instead of moving them

### LINKED ISSUE #1410: Using a chosen-element always fires a horizontal page scroll to the left

### LINKED ISSUE #1413: Use "top: -9999px" to hide elements instead of "left: -9999px" please!

### LINKED ISSUE #1686: onFocus causes auto scroll to x=0 when search is hidden
When search box is hidden, and a chosen select is clicked(only on focus), the page auto-scrolls to the hidden search input box, which should not be the case.

Sample:-

http://jsfiddle.net/2bVBN/

### LINKED ISSUE #2185: Extends beyond height of page after selecting on IE
Dear,

I'm facing an issue: the page extends beyond height after selecting on web browser IE, please take a look in the source code below:

http://jsfiddle.net/cwd7w3rm/4/

Could you please help to give me solutions to solve this issue.

Many Thanks,
Phu Dang.

### LINKED ISSUE #2402: Chosen multiple select dropdown in IE causing huge blank space below the page ( only in IE)
image
I am using chosen for multiple select , with dropdown values from angular controller. Everything is fine in Mozilla and chrome.. But in IE (I am on IE 10,but I think this is present in IE 11 as well), the functionality is working, but after selecting an option from dropdown, the browser scrollbar becomes so small and large blank space is produced at bottom of page.. This is kind of weird, it sometimes becomes normal after a click outside or when page is scrolled.. But I guess this is a bug with chosen+ IE...
Please try this on a page with a browser scroll-bar  ( ie. page which has content more than default browser window size), which has chosen multi-select dropdown...
Please tell if this is a known issue or any workarounds for this..

### LINKED ISSUE #2445: Chosen select drop down get too long when all the options are removed
In IE (I've only tried IE 11), for a chosen select with lots of options, if the drop down is clicked once and then the items are cleared from the list, the updated empty drop-down becomes massively long.

http://jsfiddle.net/3op4gmmh/8/

This is only seen when one of the divs where select element is nested has width specified specified in non-percentage value: 

    .outerdiv {
        width: 25em;
    }
or

    .outerdiv {
        width: 250px;
    }

However, if I change width of all the divs, where select is nested, to a percentage value, the issue is not seen.

    .outerdiv {
        width: 90%;
    }

The problem is my select is actually in a bootstrap's modal where I don't have control on the widths of the parent elements.

koenpunt:
I have tested this in the following browsers:

| Windows 7 | macOS Sierra |
 --- | ---
Latest Chrome | Latest Chrome
Latest Firefox | Latest Firefox
IE8 | Latest Safari
IE9 | 

There's only 1 issue on iOS, and that's that the cursor is visible, even when clipped:
image

fixes #1413
fixes #1410
fixes #1686
fixes #2402
fixes #2445
fixes #2185

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -6,10 +6,6 @@
   <link rel="stylesheet" href="docsupport/style.css">
   <link rel="stylesheet" href="docsupport/prism.css">
   <link rel="stylesheet" href="chosen.css">
-  <style type="text/css" media="all">
-    /* fix rtl for demo */
-    .chosen-rtl .chosen-drop { left: -9000px; }
-  </style>
 </head>
 <body>
   <form>
--- a/public/index.proto.html
+++ b/public/index.proto.html
@@ -6,10 +6,6 @@
   <link rel="stylesheet" href="docsupport/style.css">
   <link rel="stylesheet" href="docsupport/prism.css">
   <link rel="stylesheet" href="chosen.css">
-  <style type="text/css" media="all">
-    /* fix rtl for demo */
-    .chosen-rtl .chosen-drop { left: -9000px; }
-  </style>
 </head>
 <body>
   <div id="container">
--- a/sass/chosen.scss
+++ b/sass/chosen.scss
@@ -14,16 +14,16 @@ $chosen-sprite-retina: url('chosen-sprite@2x.png') !default;
   .chosen-drop {
     position: absolute;
     top: 100%;
-    left: -9999px;
     z-index: 1010;
     width: 100%;
     border: 1px solid #aaa;
     border-top: 0;
     background: #fff;
     box-shadow: 0 4px 5px rgba(#000,.15);
+    clip: rect(0,0,0,0);
   }
   &.chosen-with-drop .chosen-drop {
-    left: 0;
+    clip: auto;
   }
   a{
     cursor: pointer;
--- a/sass/chosen.scss
+++ b/sass/chosen.scss
@@ -373,10 +373,6 @@ $chosen-sprite-retina: url('chosen-sprite@2x.png') !default;
       }
     }
   }
-  &.chosen-container-single-nosearch .chosen-search,
-  .chosen-drop {
-    left: 9999px;
-  }
   &.chosen-container-single .chosen-results {
     margin: 0 0 4px 4px;
     padding: 0 4px 0 0;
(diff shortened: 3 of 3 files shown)


koenpunt:
@harvesthq/chosen-developers @stof as this solves quite a few issues, I'd like to see this merged, so please have a look :)

mlettini:
I forget our iOS stance, but I think it's still "Chosen is not optimal on iOS", so I'm +1 on this.

koenpunt:
Why do we use the search input for capturing focus? Can't we just add a tabindex to the container?

kevin-brown:
> Why do we use the search input for capturing focus?

I'm going to place my bets on it involving browsers not properly sending keystrokes when it was sent to the container, so a search input was used so you can properly preserve the keystrokes that people use when focused on the input.

_Note that this is why Select2 did the exact same thing before, and the browser situation hasn't improved._

koenpunt:
Everybody agrees with @mlettini?

jim-at-miramontes:
Sorry if I'm muddying things up here, but I'm not quite following the discussion flow.  I'm running v. 1.7.0 of chosen, which, after checking the code, seems to have the file updates posted as part of this page, but I'm still seeing the cursor on iPad.  Should I be (or not be)?  Is the point of the discussion that it's being punted for now, and that a real resolution of the issue will be coming later?  Thanks!

tjschuck:
@jim-at-miramontes It seems like the PR was to fix the issue on desktop browsers, and the iPad issue was punted on.

PR #2733, opened 2016-10-20
General touchup

koenpunt:
- `search_field_disabled` has now less code duplication
- set tabindex on original variable instead of through the jQuery instance
- clean up some functions unused arguments
- replace custom function scoping with fat arrow functions at their definitions  
  So instead of:
  
  ``` coffeescript
  some_method: (evt) ->
    # ...
  
  $('.el').bind 'click', (evt) => this.some_method(evt)
  ```
  
  It's now:
  
  ``` coffeescript
  some_method: (evt) =>
    # ...
  
  $('.el').bind 'click', this.some_method
  ```
  
  It might be necessary to restore some of the `return` statements at the end of the functions, although I have not seen issue yet without them.

### DIFF
--- a/coffee/chosen.jquery.coffee
+++ b/coffee/chosen.jquery.coffee
@@ -110,7 +110,7 @@ class Chosen extends AbstractChosen
     @form_field_label.unbind 'click.chosen' if @form_field_label.length > 0
 
     if @search_field[0].tabIndex
-      @form_field_jq[0].tabIndex = @search_field[0].tabIndex
+      @form_field.tabIndex = @search_field[0].tabIndex
 
     @container.remove()
     @form_field_jq.removeData('chosen')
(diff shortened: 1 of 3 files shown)


coffee/chosen.jquery.coffee
```diff
@@ -104,7 +104,7 @@ class Chosen extends AbstractChosen
       @container.bind 'click.chosen', (evt) -> evt.preventDefault(); return # gobble click of anchor
 
   destroy: ->
-    $(@container[0].ownerDocument).unbind 'click.chosen', @click_test_action
+    $(@container[0].ownerDocument).unbind 'click.chosen', this.test_active_click
```
marcandre: Why not write `@test_active_click`?

coffee/chosen.jquery.coffee
```diff
@@ -71,46 +71,46 @@ class Chosen extends AbstractChosen
...
-    @form_field_jq.bind "chosen:close.chosen", (evt) => this.close_field(evt); return
-
-    @search_field.bind 'blur.chosen', (evt) => this.input_blur(evt); return
-    @search_field.bind 'keyup.chosen', (evt) => this.keyup_checker(evt); return
-    @search_field.bind 'keydown.chosen', (evt) => this.keydown_checker(evt); return
-    @search_field.bind 'focus.chosen', (evt) => this.input_focus(evt); return
-    @search_field.bind 'cut.chosen', (evt) => this.clipboard_event_checker(evt); return
-    @search_field.bind 'paste.chosen', (evt) => this.clipboard_event_checker(evt); return
+    @container.bind 'touchstart.chosen', this.container_mousedown
+    @container.bind 'touchend.chosen', this.container_mouseup
```
stof: we must be sure that these methods can never return `false` by mistake, as this would stop propagation and prevent default.
And the automatic return of CoffeeScript could hide this from us.

PR #2736, opened 2016-10-21
Made "no_result_test" be available templating

ginlime:
With "%s" string in "no_result_test", templating available.

### DIFF
--- a/coffee/chosen.jquery.coffee
+++ b/coffee/chosen.jquery.coffee
@@ -428,7 +428,8 @@ class Chosen extends AbstractChosen
     this.result_do_highlight do_high if do_high?
 
   no_results: (terms) ->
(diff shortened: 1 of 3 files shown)


koenpunt:
Thanks for your contribution. You're not the first to propose this.

However, in an effort to try and get Chosen's repository under control and to best serve the people who use the tool, we're closing pull requests that add new functionality. For the time being, we're only merging bug fixes and functionality changes that are inline with Chosen's current scope. This doesn't mean we won't ever consider this functionality, but it does mean that we're being realistic about the time constraints the project maintainers have to give to this project.

ginlime:
Oh, I see.
But please bear in mind that your software is used all over the world.
At least, i18n or templating is needed, I think (just only for me...)

PR #2739, opened 2016-10-22
Fixed bug in 'destroy'

CliffS:
When destroying a field, results_build() is called but breaks on
@form_field.options.length as  @form_field.options is undefined.
The simplest fix is to set @disable_search on destroy which matters
not as the instance is being destroyed.


koenpunt:
Thanks for your contribution. It's not completely clear to me what bug you're fixing, because I called trying `chosen('destroy')` in multiple ways, without any errors.

If you can provide a test case, preferably a [fiddle](https://jsfiddle.net), that would be great!

PR #2761, opened 2016-12-05
Implement the ability to search with true regexes

mcclurem:
Many times users want the ability to search via regex instead of simply searching for literal strings with special characters. This PR implements a new option flag "escape_special_characters" which is defaulted to true to match old behavior. Docs and coffeescript updated and pass grunt


mcclurem:
@koenpunt any chance you have some time to review this?

PR #2596, opened 2016-04-27
Accessibility updates #264 - Add aria attributes to better support screen readers and keyboard navigation

cooperfellows:
### Summary

This adds a good chunk of ARIA attributes and other related attributes to the chosen elements to move towards making Chosen more accessible. 

**ARIA Labels and Other Helpful Attributes**
- Adds the following aria elements to the search input box
  - [role="combobox"](https://www.w3.org/TR/wai-aria/roles#combobox)
  - [aria-haspopup](https://www.w3.org/TR/wai-aria/states_and_properties#aria-haspopup) (implicit when using the combobox role)
  - [aria-expanded](https://www.w3.org/TR/wai-aria/states_and_properties#aria-expanded)
  - [aria-activedescendant="id_of_highlighted_option"](https://www.w3.org/TR/wai-aria/states_and_properties#aria-activedescendant)
  - [aria-owns="id_of_list_of_options"](https://www.w3.org/TR/wai-aria/states_and_properties#aria-owns)
  - [aria-autocomplete="list"](https://www.w3.org/TR/wai-aria/states_and_properties#aria-autocomplete)
  - [aria-labeledby="id_of_field_label"](https://www.w3.org/TR/wai-aria/states_and_properties#aria-labelledby)
- Adds the following attributes to the list of options
  - id
  - [role="listbox"](https://www.w3.org/TR/wai-aria/roles#listbox)
- Adds the following attributes to each individual option in the list of options
  - id
  - [role="option"](https://www.w3.org/TR/wai-aria/roles#option)
  - [aria-busy](https://www.w3.org/TR/wai-aria/states_and_properties#aria-busy)

**Managing State**
- Managing the aria-expanded attribute
  - In order to indicate when the search results should be relevant to assitive technology, we need to toggle the aria-expanded attribute as fields are activated/deactivated.
  - The easiest way to do this (AFAIK) is to adjust the attribute during the `close_field` and `activate_field` methods.
  - A simple switch from true to false or false to true in each of these methods is sufficient enough to keep the state updated

A complete description of the changes made, with the reasoning behind each attribute is available here: 
- https://github.com/harvesthq/chosen/issues/264#issuecomment-215124540
### References
- [Accessibility issues with chosen #264](https://github.com/harvesthq/chosen/issues/264)

PR #2764, opened 2016-12-22
A11y updates

latagore:
### Summary

This builds on pull request #2596, addressing the last comments from @KITSKevinBonett and @cooperfellows (see https://github.com/harvesthq/chosen/issues/264#issuecomment-230268619, https://github.com/harvesthq/chosen/issues/264#issuecomment-232185234 and https://github.com/harvesthq/chosen/issues/264#issuecomment-232286392). It looks like this issue or code hasn't been updated for a couple months now, so hopefully these contributions help.

I added:
- `aria-hidden=false` to the dropdown when it is loaded. The aria-hidden attribute is toggled when the dropdown is opened and closed.
- the `hidden` attribute to the native `<select>` element, to indicate that it's hidden to all users.
- `aria-selected` attributes to selected results, to indicate to assistive technologies that some entries are selected.
- surrounded the highlighted result text with a `<strong>` element so there is an indicator which element is highlighted when CSS is disabled.

### References

Based on the pull request in #2596.  Addresses issue #264.


latagore:
Discovered a bug where highlighting rows doesn't work correctly. I'm closing this and will re-open when I I fix it.

latagore:
Will reopen this pull request when I'm more confident it's stable. Currently clicking on the results text in Chrome does not work because of the newly added `<strong>` element.

latagore:
I removed the `<strong>` highlighting because, for some reason, Chrome will not fire `mouseup` or `click` events to the newly added `<strong>` element, which essentially breaks the widget. If anyone can provide any feedback, that would be great.

I also don't know the case for CSS-disabled browsing is very strong (pun not intended), since elements with the `hidden` attribute, `aria-hidden" attribute, CSS property `display: none` are displayed anyways, which can make for very confusing navigation. The highlighted row is indicated with the `aria-activedescendant` property, which I think is sufficient.

trickydisco78:
Has this made it into a release yet?

stof:
as it is not merged, no.

latagore:
I don't think I should keep people's hopes up with this pull request. When I created this, I wasn't confident in the quality of the code I wrote. I still don't feel that I understand Chosen well enough to resolve the conflicts and integrate it well with the existing code.

Issue #933, opened 2012-12-07
Use display:none for hiding lists below a certain length


pfiller:
For lists below N length, there are only minimal performance gains to be seen using `left: -999px` vs `display:none`. Chosen should toggle between the two modes based on the number of elements that are in the list. Alternatively, an option could be set to always use one way or another.

Something like this:

``` javascript
visually_show_results: ->
  if @results_data.length > 100 || @hide_results_with_position_offscreen
    @dropdown.css {"top":  dd_top + "px", "left":0}
  else
    @dropdown.css({"top":  dd_top + "px"}).show()
```

In addition to solving some browser height issues people have reported in the past, this would also improve things like #932 - where searching within a page returns strange results.

Issue #888, opened 2012-11-02
Set dropdown as display: none when hidden


johnsardine:
I was having issues with unknown space in the bottom of my page when I found out that it was a hidden chosen dropdown. I suggest adding display: none when the dropdown is hidden.

Demonstration of the problem.

**Before:**

Closed: http://cl.ly/Kcs6
Open: http://cl.ly/KdD5

As you can see, chosen was taking unecessary space.

I added display: none (and display: block to make it visible)

**After:**

Closed: http://cl.ly/KcEF
Open: http://cl.ly/Kczf

Now it only takes up space when necessary, which I dont mind, since it's not a system element and does not adapt to the available space like default elements.

I don't submit a pull request because I don't have the necessary dependencies or know-how, and since this is such a small fix, It won't take much time to change if anyone feels this should be changed.

PR #2780, opened 2017-02-22
Replace deprecated CSS clip property

bingxie:
### Summary

Previously, chosen used CSS clip property to hide the dropdown. There are two reasons for this PR to replace it. 

1. CSS clip property is deprecated. [MDN CSS clip](https://developer.mozilla.org/en/docs/Web/CSS/clip)
2. In IE9/10 the clip property causes the dropdown does not hide properly. 

This change has been tested in a big web application.

Please double-check that:

### References

If your pull request is in reference to one or more open GitHub issues, please mention them here to keep the conversations linked together.

### DIFF
--- a/sass/chosen.scss
+++ b/sass/chosen.scss
@@ -20,10 +20,10 @@ $chosen-sprite-retina: url('chosen-sprite@2x.png') !default;
     border-top: 0;
     background: #fff;
     box-shadow: 0 4px 5px rgba(#000,.15);
-    clip: rect(0,0,0,0);
+    display: none;
   }
   &.chosen-with-drop .chosen-drop {
-    clip: auto;
+    display: block;
   }
   a{
     cursor: pointer;
@@ -136,8 +136,7 @@ $chosen-sprite-retina: url('chosen-sprite@2x.png') !default;
     background-clip: padding-box;
   }
   &.chosen-container-single-nosearch .chosen-search {
-    position: absolute;
-    clip: rect(0,0,0,0);
+    display: none;
   }
 }
 /* @end */


sprankhub:
:+1: Thanks!

I had issues with that just now as well. `display: block/none` seems to work (better) for me, too.

sass/chosen.scss
```diff
@@ -137,7 +137,7 @@ $chosen-sprite-retina: url('chosen-sprite@2x.png') !default;
   }
   &.chosen-container-single-nosearch .chosen-search {
     position: absolute;
```
mlettini: This property is no longer necessary. It's a hold over from when we used to do `left: -9999px; position: absolute;` to hide the search box.

bingxie: That's very good point, thank. I will remove it.

koenpunt:
Display none/block didn't perform well on large lists, so then absolute positioning was introduced, moving elements out of the viewport (i.e. `left: -10000`) which caused issues with focusing. 

So please be cautious with merging, because clip actually performed best of all options.

sass/chosen.scss
koenpunt: Not exactly, clip also only works on absolute positioned elements;)

sprankhub:
@koenpunt could you elaborate what you mean with "did not perform well on lare lists"? Do you really mean performance or something else?

mlettini:
@koenpunt Also curious about the performance aspect. And if display none/block is not going to perform well, would you suggest we go back to moving it out of the viewport with `left: -9999px`?

bingxie:
What I can provide is in our application, I tested on very large list, like hundreds of items. I have not seen any performance issues. I also would like to know more about other related issues.

westonganger:
This PR fixes problems I was experiencing in Internet Explorer 11 as well.

koenpunt:
Never seen the performance issues myself, and don't know if they still exist in modern browsers. But I know that @pfiller brought it up some times, e.g.

https://github.com/harvesthq/chosen/issues/888#issuecomment-10820094

https://github.com/harvesthq/chosen/issues/933

But that concern might have not been relevant anymore since https://github.com/harvesthq/chosen/pull/1339

And yes I know, these are all pretty old issues.

Anyway, if @pfiller could chime in about this matter that would be great

koenpunt:
Maybe the question we need to ask ourselves is; do select options have to be accessible? Not that Chosen is really accesible right know, but at least the options are being read by screen readers and search engines.

How html5 boilerplate does it might be a solution too:

https://github.com/h5bp/html5-boilerplate/blob/master/src/css/main.css#L128

lusever:
With these changes, tabbing does not work on the control.

tjschuck:
Cross-linking [this comment](https://github.com/harvesthq/chosen/issues/2816#issuecomment-341441917) about this apparently being truly broken in Chrome 64, and:

> According to their release schedule, looks like Chrome 64 is scheduled to hit stable on Jan 23rd, 2018, at which point we’ll be broken.

/cc @adunkman @mlettini @koenpunt

mlettini:
So to summarize where we're at right now:

* Using `clip` is deprecated and will break Chosen in Chrome 64 on Jan 23rd, 2018.
* Using `display: none/block` results in not being able to tab to a single Chosen.
* Using `position: absolute` and pushing the dropdown off the page will result in all the issues linked to [here](https://github.com/harvesthq/chosen/pull/2727).

So we need to find another method of hiding the Chosen drop until it gets focus. [HTML5 Boilerplate](https://github.com/h5bp/html5-boilerplate/blob/master/src/css/main.css#L128) will also need to update, and I wonder if there is any current conversation on this topic happening there. No obvious solutions jump to my mind at the moment...

adunkman:
Thanks for the summary @mlettini!

I think the core issue stems from us using a text input inside of the drop container to receive focus events. This makes sense when that text input is always visible (when `<select multiple>`) but not when it’s hidden. 

I’ve toyed around a bit with using two inputs, one to catch focus events (which lives outside of the chosen drop container) and the second being the text input that currently exists. This allows us to just use `display: none` on the drop container when appropriate, but still catch focus events. 

I don’t have a working pull at the moment, but I’m mostly convinced it’s the right way to go — unless y’all have better ideas!

adunkman:
Replaced by #2939.

PR #2785, opened 2017-03-07
Inform previous naming conventions

diegofaria:
### Summary

I spent a little decent amount of time to find that "chosen" was "liszt".
I don't know if this is the best choice, but I think that would be great to be more clear about it.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -1336,7 +1336,7 @@
           </li>
           <li>
             <h3>Updating Chosen Dynamically</h3>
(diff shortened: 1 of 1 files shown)


tjschuck:
Eh, those names were removed 4 years ago.  I think it's okay to move on — we don't get very many people having this old-API information as a problem these days.

I'm assuming you found some reference to `liszt` via StackOverflow or something?  I think your best bet would just be to leave a comment there (or update the answer, if you have the reputation).

Thanks for the PR, though!  At least the existence of this PR might help a future searcher.

Issue #2805, opened 2017-04-03
Version 1.7.0 - search bug with cyrillic


Jeckerson:
By typing any cyrillic characters in search input - the output is the **last character of option value**.

This affects only cyrillic characters.

koenpunt:
I don't think this is an issue with 1.7.0 alone, but also applies to previous versions?

fuzegit:
This bug has appeared since 1.7.0 version. In 1.6.2 version everything works perfectly.

eesau:
This bug is not limited to cyrillic characters, any non-ascii character in the option text sees to trigger it.
1.6.2 and before work fine.
See https://jsfiddle.net/bpnpzgh8/14/

geminorum:
I too have this problem with Persian.

Sogl:
Same here.

List:
![image](https://cloud.githubusercontent.com/assets/7556290/26380022/d38c5c26-4067-11e7-879d-d2506c09ae83.png)

With search:
![image](https://cloud.githubusercontent.com/assets/7556290/26379907/4bf23b0a-4067-11e7-8be6-51892e8aac46.png)

When will be fixed?

msom:
https://github.com/harvesthq/chosen/pull/2811 seems to work:
![screen shot 2017-09-20 at 09 05 57](https://user-images.githubusercontent.com/5045627/30630966-f3f5e472-9de2-11e7-8490-c66e2357b63e.png)

adunkman:
This has been fixed in versions 1.8.1 and above.

PR #2811, opened 2017-04-16
Fix match highlighting, if multi-byte character is next to a word boundary

kgilden:
## Before
old

## After
new

## Explanation

Hi, I made some changes to the highlighting regex so that it would for search patterns, where the first character is non-ASCII (e.g. "š").

Previously, for patterns such as `š` chosen would find the correct options (e.g. `foo & šoo`), but the highlighting code would cause the options to be rendered as a single character (e.g. `f`).

The solution is not to use the `\b` (only works for ASCII characters) and explicitly check for either beginning of a string or a whitespace character. I then had to use capturing groups + `String.prototype.indexOf`, because the whitespace character is also included in the entire matched string.

More details can be found [here](http://stackoverflow.com/a/10590516) and [here](http://stackoverflow.com/a/2449892).

### DIFF
--- a/coffee/lib/abstract-chosen.coffee
+++ b/coffee/lib/abstract-chosen.coffee
@@ -193,7 +193,7 @@ class AbstractChosen
 
           if option.search_match
             if searchText.length
-              startpos = option.search_text.search highlightRegex
+              startpos = option.search_text.indexOf highlightRegex.exec(option.search_text)[1]
               text = option.search_text.substr(0, startpos + searchText.length) + '</em>' + option.search_text.substr(startpos + searchText.length)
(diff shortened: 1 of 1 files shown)


kgilden:
Any comments on this?

stof:
word delimiters are not only spaces. Punctuation also serves as delimiter for instance.

geminorum:
related: #2805

kgilden:
@adunkman I see that this has been fixed, should I close this PR?

adunkman:
Nice! Thanks for bringing it up — yes, it looks like this is fixed now. ✨ 

Happy Friday!

PR #2844, opened 2017-06-22
Fix html entity in search results

satchmorun:
Don't search escaped HTML text

Before this change, if you tried to search for "&" to match "Stuff & Nonsense", what we would do is html-escape _both_ things and compare the escaped strings.

This works fine in the case where you just type "&". However, if you type "mp", that will match the html entity "&amp;" in the middle of the html-escaped text we're comparing it to. Which is incorrect.

Additionally, we highlight our "matched" result. When we do, we break up the entity so that the resulting HTML looks like:

    Stuff &a<em>mp</em>; Nonsense

And you get to see what basically looks like garbage on the screen:

screen garbage

The meat of this PR and this PR description comes from [this commit](https://github.com/harvesthq/chosen/commit/0adf2d5bf275dcbdfd72185959a4782158fe36a5). The subsequent commits are just refactorings and tidyings.

PR made in a pairing session with @braddunbar.

### DIFF
--- a/coffee/lib/abstract-chosen.coffee
+++ b/coffee/lib/abstract-chosen.coffee
@@ -185,7 +185,7 @@ class AbstractChosen
           results += 1 if results_group.active_options is 0 and results_group.search_match
           results_group.active_options += 1
 
-        option.search_text = if option.group then option.label else option.html
+        option.search_text = if option.group then option.label else option.text
 
         unless option.group and not @group_search
           search_match = this.search_string_match(option.search_text, regex)
@@ -193,11 +193,14 @@ class AbstractChosen
 
           results += 1 if option.search_match and not option.group
 
+
           if option.search_match
             if searchText.length
               startpos = search_match.index
-              text = option.search_text.substr(0, startpos + searchText.length) + '</em>' + option.search_text.substr(startpos + searchText.length)
-              option.search_text = text.substr(0, startpos) + '<em>' + text.substr(startpos)
+              prefix = option.search_text.slice(0, startpos)
+              fix    = option.search_text.slice(startpos, startpos + query.length)
+              suffix = option.search_text.slice(startpos + query.length)
+              option.search_text = "#{this.escape_html(prefix)}<em>#{this.escape_html(fix)}</em>#{this.escape_html(suffix)}"
 
             results_group.group_match = true if results_group?
 
--- a/coffee/lib/abstract-chosen.coffee
+++ b/coffee/lib/abstract-chosen.coffee
@@ -116,7 +116,7 @@ class AbstractChosen
     option_el.className = classes.join(" ")
     option_el.style.cssText = option.style
     option_el.setAttribute("data-option-array-index", option.array_index)
-    option_el.innerHTML = option.search_text
+    option_el.innerHTML = option.highlighted_html or option.html
     option_el.title = option.title if option.title
 
     this.outerHTML(option_el)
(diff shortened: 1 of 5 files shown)


coffee/lib/abstract-chosen.coffee
```diff
@@ -185,17 +186,19 @@ class AbstractChosen
...
 
           if option.search_match
-            if searchText.length
-              startpos = option.search_text.search highlightRegex
-              text = option.search_text.substr(0, startpos + searchText.length) + '</em>' + option.search_text.substr(startpos + searchText.length)
-              option.search_text = text.substr(0, startpos) + '<em>' + text.substr(startpos)
+            if query.length
+              startpos = text.search highlightRegex
+              prefix = text.slice(0, startpos)
+              fix    = text.slice(startpos, startpos + query.length)
```
adunkman: A+ variable name.

tjschuck:
See also:

* #1150
* #1638
* #2254
* #2548

Plus like a scrillion other issues and PRs linked off of those.

satchmorun:
I'm sorry – I had forgotten about this PR.

Just got this up-to-speed and mergeable with https://github.com/harvesthq/chosen/pull/2844/commits/800b7c901e96d5f700c4ea773da79ac8304d7f2d.

> There has been changes (and issues) in this department quite some times, so we should be certain that this won't reintroduce one of the earlier problems.

@koenpunt I feel pretty certain that we aren't reintroducing the earlier problems. The [first commit](https://github.com/harvesthq/chosen/pull/2844/commits/0adf2d5bf275dcbdfd72185959a4782158fe36a5) in this PR is what prevents the same problem that was fixed in #2254. The main problem with the prior attempts was not escaping the HTML as it's built up for the highlighted result, and here we're very explicit about it: https://github.com/harvesthq/chosen/pull/2844/commits/0adf2d5bf275dcbdfd72185959a4782158fe36a5#diff-a3162852f1360124c43007c8130261dcR201

/cc @adunkman @tjschuck

coffee/lib/abstract-chosen.coffee
```diff
@@ -185,19 +185,21 @@ class AbstractChosen
...
           if option.search_match
-            if searchText.length
+            if query.length
               startpos = search_match.index
-              text = option.search_text.substr(0, startpos + searchText.length) + '</em>' + option.search_text.substr(startpos + searchText.length)
-              option.search_text = text.substr(0, startpos) + '<em>' + text.substr(startpos)
+              prefix = text.slice(0, startpos)
+              fix    = text.slice(startpos, startpos + query.length)
+              suffix = text.slice(startpos + query.length)
+              option.highlighted_html = "#{@escape_html(prefix)}<em>#{@escape_html(fix)}</em>#{@escape_html(suffix)}"
```
koenpunt: For functions we use `this` instead of `@`, so please use `this.escape_html(prefix)`

koenpunt:
Probably not a change that you can add some tests for this? 😅

coffee/lib/abstract-chosen.coffee
satchmorun: Updated in https://github.com/harvesthq/chosen/pull/2844/commits/5940e3b88213fc606a17001437f672b2e7d4c15d.

satchmorun:
> Probably not a change that you can add some tests for this? 😅

I think maybe I can. I'll try and see what happens.

satchmorun:
@koenpunt we've now got regression tests for this! https://github.com/harvesthq/chosen/pull/2844/commits/d4715b1a72b90039fa971e011e6b68e96306255f

Bing bang boom.

tjschuck:
@satchmorun Can you squash this down to one (or a few) logical commits?

satchmorun:
@tjschuck done and done

koenpunt:
> Bing bang boom.

Great! :shipit:

PR #2859, opened 2017-07-31
Update dev dependencies to use uptodate packages

stof:
All dependencies used in the build system were totally outdated (and may not work well on recent versions of node). This updates them.


public/index.html
```diff
@@ -7,7 +7,7 @@
   <link rel="stylesheet" href="docsupport/prism.css">
   <link rel="stylesheet" href="chosen.css">
 
-  <meta http-equiv="Content-Security-Policy" content="default-src 'self'; script-src 'self' https://ajax.googleapis.com; style-src 'self'; img-src 'self' data:">
+  <meta http-equiv="Content-Security-Policy" content="default-src &apos;self&apos;; script-src &apos;self&apos; https://ajax.googleapis.com; style-src &apos;self&apos;; img-src &apos;self&apos; data:">
```
tjschuck: Is this right?  Looking around, it seems like most `"Content-Security-Policy"` examples use the single-quote-inside-double-quote thing.

stof: well, escaping is not *required*, but it is allowed (this is something happening at the HTML level, so the final value has single quotes after parsing).

Escaping the single quote is the result of running dom_munger as part of the `prep-release` task (the fact that it was not escaped yet is because this change was reverted manually when doing the 1.7.0 release)

tjschuck: It's weird, but it seems to work, so okay ¯\\\_(ツ)_/¯

PR #2879, opened 2017-08-31
Fix escaping in group names, no results message.

adunkman:
@harvesthq/chosen-developers 

Surprisingly, I don’t think I found any issues in this repo for this one — but we’re incorrectly escaping group names and the no results message in some cases: 

- Group name is not properly escaped when it contains a highlighted search term.
- The user-entered text is never properly escaped in the no results message.

| Before this PR | After this PR | Status |
|---------------|-------------|--------|
|image|image|(no issue)|
|image|image|Fixed|
|image|image|(no issue)|
|image|image|Fixed|
|image|image|(no issue)|

Test cases included.

### DIFF
--- a/spec/proto/searching.spec.coffee
+++ b/spec/proto/searching.spec.coffee
@@ -72,3 +72,27 @@ describe "Searching", ->
 
     expect(div.select(".group-result").length).toBe(1)
     expect(div.down(".group-result").innerHTML).toBe("<em>A</em> &amp; B")
+
+  it "renders no results message correctly when it contains characters that require HTML encoding", ->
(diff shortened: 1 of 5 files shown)


tjschuck:
That before/after table is a thing of beauty, @adunkman 😻

spec/proto/searching.spec.coffee
```diff
@@ -28,3 +28,72 @@ describe "Searching", ->
 
     results = div.select(".active-result")
     expect(results.length).toBe(0)
+
+
```
tjschuck: ✂️

PR #2880, opened 2017-08-31
Only trim input_width to container_width when calculable.

### LINKED ISSUE #2480: Default text for multiple select is being truncated
I'm using bootstrap and this is my html code:

``` html
<div class="form-group">
<select id="copyproposalpageform-proposal_ids" class="chosen" name="CopyProposalPageForm[proposal_ids][]" multiple="multiple" size="4">

</select>
</div>
```

``` javascript
$('.chosen').chosen({
    width: '100%'
});
```

The default text is being truncated
screen shot 2015-12-02 at 11 23 17 am
but after I clicked to the dropdown and click away, it back to normal

adunkman:
@harvesthq/chosen-developers to fix #2480.

This PR adjusts the logic used to calculate the width of the text input when in multi-select mode. 

If Chosen is not visible when initialized (it’s within a hidden container), the calculated width of Chosen’s container is `0`. This means that we’re setting the search input width to an initial value of `-10px`, which is nonsense — resulting in a fallback to it’s default width in CSS (25px). 

This, of course, is adjusted after Chosen is interacted with, since it is now visible and the container width can be correctly calculated. 

Since we don’t know when Chosen becomes visible, I propose we adjust the default width for the search input to be the full calculated width, and ignore the calculated container_width if it results in zero. The text doesn’t actually overflow outside of the container — the only downside is that a few pixels of space isn’t present. 

Here’s a GIF of Chosen (which was rendered off-screen initially) being opened and closed. Before, the width of the search input was not contained to the container, and after it is:

screencast

If you didn’t catch it, look for the tail end of the “a” character. This is _significantly better_ than what’s currently happening — the input is rendered at 25px wide:

|  | Before this PR | After this PR |
|--|---------------|--------------|
|Before Opening|image|image|
|After Opening|image|image|

### DIFF
--- a/coffee/chosen.jquery.coffee
+++ b/coffee/chosen.jquery.coffee
@@ -500,9 +500,8 @@ class Chosen extends AbstractChosen
     width = div.width() + 25
     div.remove()
 
-    container_width = @container.outerWidth()
-
-    width = Math.min(container_width - 10, width)
+    if @container.is(':visible')
+      width = Math.min(@container.outerWidth() - 10, width)
 
     @search_field.width(width)
 
--- a/coffee/chosen.proto.coffee
+++ b/coffee/chosen.proto.coffee
@@ -505,9 +505,8 @@ class @Chosen extends AbstractChosen
     width = div.measure('width') + 25
     div.remove()
 
-    container_width = @container.getWidth()
-
-    width = Math.min(container_width - 10, width)
+    if container_width = @container.getWidth()
+      width = Math.min(container_width - 10, width)
 
     @search_field.setStyle(width: width + 'px')
 


coffee/chosen.jquery.coffee
```diff
@@ -500,9 +500,8 @@ class Chosen extends AbstractChosen
     width = div.width() + 25
     div.remove()
 
-    container_width = @container.outerWidth()
-
-    width = Math.min(container_width - 10, width)
+    if @container.is(':visible')
+      width = Math.min(@container.outerWidth() - 10, width)
```
satchmorun: @adunkman and I paired on this, because I couldn't reproduce the issue in my jquery HTML page.

While working on it, we discovered a fun issue with jquery's [outerWidth method](http://api.jquery.com/outerWidth/). The short of it is that if the element is hidden, one should not trust jquery to do anything sensible. For example, with a `style` of `width: 98%` (on a non-visible element), it will return `98`. _Le sigh_.

So we needed to do a different thing in the jQuery version than in the prototype version.

Go early 2000s!

PR #2881, opened 2017-09-05
Escape optgroup label when appending to HTML.

adunkman:
@harvesthq/chosen-developers 

Hot on the tails of the 1.8.1 release (and as a follow-up to #2879), this PR fixes a cross-site scripting vulnerability discovered when applying the 1.8.1 release to Harvest. 

When an optgroup’s label is not highlighted, we’re using the unescaped version of the `label` attribute — but we’re safe when the highlighting is added, since we’re [escaping each HTML fragment](https://github.com/harvesthq/chosen/blob/908b0bfc819ef407532c7bf289959045ea3c0b34/coffee/lib/abstract-chosen.coffee#L203) when building the highlighted string.

Included is a test which asserts HTML tags are not rendered into HTML unescaped in this case.

### DIFF
--- a/coffee/lib/abstract-chosen.coffee
+++ b/coffee/lib/abstract-chosen.coffee
@@ -131,7 +131,7 @@ class AbstractChosen
 
     group_el = document.createElement("li")
     group_el.className = classes.join(" ")
-    group_el.innerHTML = group.highlighted_html or group.label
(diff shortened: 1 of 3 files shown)


garpalhares:
(Sorry my bad english)
I can not disable of the plugin boxes in mobile. Stay like this: http://prntscr.com/h0btw4

Issue #2816, opened 2017-04-28
Remove CSS clip property


mateuspv:
I noticed that chosen is using `clip` property, with is deprecated...I would like to know if is possible to remove it.

> This feature has been removed from the Web standards. Though some browsers may still support it, it is in the process of being dropped. Avoid using it and update existing code if possible; see the compatibility table at the bottom of this page to guide your decision. Be aware that this feature may cease to work at any time.

Font: https://developer.mozilla.org/en-US/docs/Web/CSS/clip

Any hint's about how to do this?

koenpunt:
There's already a PR pending for this; https://github.com/harvesthq/chosen/pull/2780

H3xx3n:
On chrome 64 (dev) does no longer works. The clip property is making the dropdown invsibile

adunkman:
Thanks for the heads-up, @H3xx3n! According to [their release schedule](https://www.chromium.org/developers/calendar), looks like Chrome 64 is scheduled to hit stable on  Jan 23rd, 2018, at which point we’ll be broken.

PR #2886, opened 2017-09-20
 * Fixed - prevent clicking on hidden results scrollbar in IE

cbarber:
Resolve IE issue where clicking on the hidden results scrollbar triggers the chosen select to display.

Steps:
* Create chosen select with enough results to show a scrollbar in the results div
* Focus and blur the chosen select at least once. Note the location of the scrollbar
* Click on the location of where scrollbar is (albeit hidden). Chosen select will display

Example gif:
https://imgur.com/a/pDhBu

The fix is to exclude "chosen-results" from the mousedown event. Does not seem to have side effects.

### DIFF
--- a/public/options.html
+++ b/public/options.html
@@ -89,7 +89,7 @@
         <tr>
           <td>group_search</td>
           <td>true</td>
-          <td>By default chosen will search labels of groups as well as options, and display all options below matching groups. Set this to <code class="language-javascript">false</code> to search only in actual options.</td>
+          <td>By default, Chosen will search group labels as well as options, and filter to show all options below matching groups. Set this to <code class="language-javascript">false</code> to search only in the options.</td>
         </tr>
         <tr>
           <td>single_backstroke_delete</td>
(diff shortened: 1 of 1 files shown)


tjschuck:
Weird!  Thanks for the GIF — that shows the issue perfectly.

What browser, browser version, OS, and OS version are you using?

cbarber:
Oops. Sorry I forgot to include those details. Windows 10 - Internet Explorer 11.608.15063.0

cbarber:
I discovered that any form field beneath the drop down was not receiving mouse click event where the scrollbar is located. I felt like adding more logic to the mouse down event method was the wrong the approach. I've updated the pull request to hide the scrollbar (IE only) when the drop down is deactived and show it when its activated. I feel like this is a safer way to handle the quirk.

tjschuck:
@cbarber Can you squash this down to one commit?  [Instructions here](http://blog.steveklabnik.com/posts/2012-11-08-how-to-squash-commits-in-a-github-pull-request).

cbarber:
@tjschuck Sorry I missed you message. Pull request has been updated.

tjschuck:
@cbarber Thanks!

@harvesthq/chosen-developers One of y'all wanna take a peek and review?

_Possible red herring, but:_ this change is happening around some of those `clip` attributes that are [apparently an issue](https://github.com/harvesthq/chosen/issues/2816) — any chance this would actually be fixed without vendor-prefixery via the [`clip` removal PR](https://github.com/harvesthq/chosen/pull/2780)?

cbarber:
Oops. I rebased to resolve the conflicts and it looks like a forced push to my fork closed the PR. When I was resolving the conflict I did notice that the clip css style that was related to the original problem had been switched to `display: none`. I don't have IE available right now to see if this resolves this but should be able to check in the next day or two.

Issue #2862, opened 2017-08-15
enable_split_word_search doesn't do anything if search_contains is true


scott-census:
Setting enable_split_word_search: false is supposed to make the search match the whole word (i.e. start from beginning). In winnow_results, what it does if false is insert a '^' at the beginning of the regex--but it only does so if search_contains is also false (code is `if (!(enable_split_word_search || search_contains)) ...`. This is the only place enable_split_word_search is referenced in the code, other than to set the internal variable.

I have to use search_contains: true because there are various non-alpha characters in my lists (like '+', ')', '-') and matches for them fail if they are supposed to be on word boundaries (I believe this is a known issue). But I would still like to match from the beginning. Seems like a bug to me that using search_contains prohibits matching from the start.

### Steps to reproduce

Tell us how to reproduce this issue.

  1. Create a list with options {enable_split_word_search: false, search_contains: true}

  2. Try to search for something that occurs within a word but also at the beginning of a word

I created two demos to show this issue, one using the current version (1.7.0) and the other a build from master as of Aug 8, because they work differently when search_contains is true, but in both cases that condition will not work with the lists I have.
[jsFiddle: Chosen 1.7.0](https://jsfiddle.net/6x6f3chw/11/)
[Plunker: Chosen master (Aug 3)](https://embed.plnkr.co/RIBsntyfHx9zE6At7jjk/)

Additionally, they both show bugs when search_contains is true. In particular, the current master does not show the first two list items ('<01M' and '<10Y') as all (in addition to demonstrating the known issues with non-alpha characters).  I haven't dug into that because I can't use that setting anyway.

### Expected behavior

Perhaps enable_split_word_search would be better named search_from_beginning, but it should add the '^' to the regex regardless of the value of search_contains (in get_search_regex, line 408).

### Actual behavior

Has no effect unless search_contains is false (which causes other problems in many cases).

### Environment

  - **1.7.0 and current master**:

### Additional information

I guess the current unreleased version is an improvement, although the addition of \b around the search term in the regex causes problems with a number of characters. However, dropping the items from the list altogether seems like an issue. In the 1.7.0 version, there is also a problem with those entries: they are displayed in the result list as only the alphabetic part ('M' and 'Y')

Mikk3lRo:
IMHO what you suggest is as counter-intuitive as the problem you experience. Only difference is that `enable_split_word_search=false` would override `search_contains=true` instead of the other way around.

[...]

PR #2896, opened 2017-10-02
Custom word boundary

Mikk3lRo:
### Summary

The default / current word boundary regex `^|\\b|\\s` is remarkably bad at actually finding word boundaries in languages that use non-unicode characters. A word boundary is basically detected after any non-ascii character (fx. ü, å, ø and æ to mention just a few - but there are MANY).

I've looked into possibilities, and unfortunately there doesn't seem to be any way to get decent word-boundary detection for anything except ascii in javascripts RegExp implementation... without either using a third-party library or including some 4k+ characters in the string.

Therefore, I don't see any way to reliably detect word boundaries with any pre-set, hardcoded regex.

Turning it into an option means that people can at least set something appropriate for their individual language and / or use case if they care about word boundaries being detected in "weird" places.

Please double-check that:

### References

First partial PR from: https://github.com/harvesthq/chosen/pull/2894

### DIFF
--- a/coffee/lib/abstract-chosen.coffee
+++ b/coffee/lib/abstract-chosen.coffee
@@ -27,6 +27,7 @@ class AbstractChosen
     @enable_split_word_search = if @options.enable_split_word_search? then @options.enable_split_word_search else true
     @group_search = if @options.group_search? then @options.group_search else true
     @search_contains = @options.search_contains || false
+    @search_word_boundary = if @options.search_word_boundary? then @options.search_word_boundary else '^|\\s|\\b'
     @single_backstroke_delete = if @options.single_backstroke_delete? then @options.single_backstroke_delete else true
     @max_selected_options = @options.max_selected_options || Infinity
     @inherit_select_classes = @options.inherit_select_classes || false
@@ -217,7 +218,7 @@ class AbstractChosen
       this.winnow_results_set_highlight()
 
   get_search_regex: (escaped_search_string) ->
-    regex_string = if @search_contains then escaped_search_string else "(^|\\s|\\b)#{escaped_search_string}[^\\s]*"
+    regex_string = if @search_contains then escaped_search_string else "(#{@search_word_boundary})#{escaped_search_string}[^\\s]*"
     regex_string = "^#{regex_string}" unless @enable_split_word_search or @search_contains
     regex_flag = if @case_sensitive_search then "" else "i"
     new RegExp(regex_string, regex_flag)
--- a/public/options.html
+++ b/public/options.html
@@ -92,6 +92,11 @@
           <td>By default, Chosen will search group labels as well as options, and filter to show all options below matching groups. Set this to <code class="language-javascript">false</code> to search only in the options.</td>
         </tr>
         <tr>
+          <td>search_word_boundary</td>
+          <td>^|\\b|\\s</td>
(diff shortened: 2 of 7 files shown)


Mikk3lRo:
I'm fairly certain this will solve at least this open issue / bug:
https://github.com/harvesthq/chosen/issues/2862

tjschuck:
@Mikk3lRo Can you squash this down to one commit?  [Instructions here](http://blog.steveklabnik.com/posts/2012-11-08-how-to-squash-commits-in-a-github-pull-request).

coffee/lib/abstract-chosen.coffee
```diff
@@ -27,6 +27,7 @@ class AbstractChosen
     @enable_split_word_search = if @options.enable_split_word_search? then @options.enable_split_word_search else true
     @group_search = if @options.group_search? then @options.group_search else true
     @search_contains = @options.search_contains || false
+    @search_word_boundary = if @options.search_word_boundary? then @options.search_word_boundary else '^|\\s|\\b'
```
koenpunt: Can’t this be just;

```coffee
@search_word_boundary = options.search_word_boundary || '^|\\s|\\b'
``` 

?

koenpunt: Also, the option should not necessarily include “search” in its name

public/options.html
```diff
@@ -87,11 +87,21 @@
...
           <td>group_search</td>
           <td>true</td>
           <td>By default, Chosen will search group labels as well as options, and filter to show all options below matching groups. Set this to <code class="language-javascript">false</code> to search only in the options.</td>
         </tr>
         <tr>
+          <td>search_word_boundary</td>
+          <td>^|\\b|\\s</td>
+          <td>By default, Chosen uses JS RegExp's built-in word boundary to detect word beginnings as well as whitespace or the beginning of the entire label. That works great for ascii-only languages, but <strong>will</strong> erroneously detect word boundaries after letters with umlauts among many, many others.<br>You can pass a string (that will be interpreted as part of a <code class="language-javascript">RegExp</code>) refined for your language and use case to correctly detect word boundaries. A (simplified) example could be <code class="language-javascript">'^|[^A-zæøåÆØÅ]'</code> for Danish.</td>
+        </tr>
+        <tr>
```
koenpunt: This is now in here twice

spec/jquery/searching.spec.coffee
```diff
@@ -247,7 +247,7 @@ describe "Searching", ->
 
     expect(div.find(".active-result").length).toBe(1)
     expect(div.find(".active-result")[0].innerHTML).toBe("oh <em>h</em>ello")
-
+
```
koenpunt: Nit; unrelated change

Mikk3lRo: Yeah... not eactly a git wiz here, and the squahing turnied out to not be as easy as the instuctions suggested due to a conflict on the options page. I'm just about to give up and start a new branch instead.

coffee/lib/abstract-chosen.coffee
Mikk3lRo: Is it not logical to prefix search related options with "search"?

And I'm pretty sure the question mark is needed, otherwise the empty string will evaluate to false and the default value will be used. The empty string is a perfectly valid value for this option... if set to an empty string Chosen works like `search_contains=true`, but allowing `enable_split_word_search` to decide whether to only match the beginning of the option - see linked issue for a relevant use case.

spec/jquery/searching.spec.coffee
Mikk3lRo: Ok the whole thing is a mess now. I have no idea how to recover. Can I start a new branch and pull request?

Mikk3lRo:
Sorry, but this is too messed up now. Some stupid conflict going around itself in circles.

PR #2899, opened 2017-10-03
describe => it

koenpunt:
The incorrect keyword was being used.


Mikk3lRo:
Uhh... Actually `describe` is not incorrect here...?

The `it`'s come below - `describe` is like a headline for several tests related to the boundary-detection...

koenpunt:
@Mikk3lRo I guess you're right, I missed the `it` below.

But I still then think these tests could be written more readable, and then probably then also become less error prone.

PR #2900, opened 2017-10-05
Skip highlight when holding the meta key.

### LINKED ISSUE #2888: List jumps to top after select with CMD
### Tell us how to reproduce this issue.

- https://harvesthq.github.io/chosen/
- use the multi select field
- select one item with the CMD button holding down
- list jumps after selection always to the top

### Expected behavior / Actual behavior

It should not jump to top? I dont know if this was happening in the old version, but I think it would be more user friendly if it wont jump to the top each time.

### Environment

  - 1.8.2 Chosen

  - Chrome, Safari and Firefox latest

  - macOS Sierra

Cheers
Hannes

braddunbar:
@harvesthq/chosen-developers 👋 

Fix #2888

When holding command and selecting an option from a multi-select, the selection area should not scroll back to the top. I'll write some tests and push those shortly!

### DIFF
--- a/coffee/chosen.jquery.coffee
+++ b/coffee/chosen.jquery.coffee
@@ -374,7 +374,7 @@ class Chosen extends AbstractChosen
         this.single_set_selected_text(this.choice_label(item))
 
       if @is_multiple && (!@hide_results_on_select || (evt.metaKey or evt.ctrlKey))
-        this.winnow_results()
+        this.winnow_results(skip_highlight: true)
       else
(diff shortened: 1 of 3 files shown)


adunkman:
Love it! That if-statement is full of 🐉… I’m a little worried about side-effects. That worrisome feeling will be qualmed by tests, I think!

adunkman:
Merged as a part of #2910 — this commit was reviewed there.

PR #2906, opened 2017-10-18
Mm 1935

rshariff-configero:
semicolon tags


tjschuck:
This seems like an accidental PR.  Closing.

If it's _not_, it's severely lacking in explanation.  Please re-open with more context.

PR #2910, opened 2017-10-25
Do not highlight first result or clear search input when holding control/meta key.

### LINKED ISSUE #2888: List jumps to top after select with CMD

adunkman:
@harvesthq/chosen-developers to fix #2888, replaces #2900.

Building on @braddunbar’s #2900, this PR skips highlighting the first result as described in that PR and also does not reset the search input. This ensures we can do searchless multiple select when holding the control/meta key: 

screencast of holding meta to select many options

This also ensures we can do multiple select while searching using the control/meta key:

screencast of searching and holding meta to select many options

This is similar to the suggestion that @koenpunt mentioned in https://github.com/harvesthq/chosen/pull/2867#issuecomment-324076160:

> … it might be relevant to leave the text in, if someone wants to select multiple options that matched their query. So maybe we should only clear the input when there are no more matching options visible?

This PR doesn’t exactly implement this behavior — it instead leaves the input when meta/control is pressed — but allows users to achieve the same thing.

I wasn’t able to implement a prototype version of the related test — `simulant` doesn’t appear to bubble events, so I can’t get the event handler to trigger correctly in prototype (either the event comes from the correct `.active-result` item and doesn’t hit the event handler which is registered on `.chosen-results`, or the event comes from the wrong item (`.chosen-results`) and hits the event handler). I spent… too much time (3:04) working on trying to get a test working for prototype — but at least we’re covered with a jQuery test.

### DIFF
--- a/coffee/chosen.jquery.coffee
+++ b/coffee/chosen.jquery.coffee
@@ -374,7 +374,7 @@ class Chosen extends AbstractChosen
         this.single_set_selected_text(this.choice_label(item))
 
       if @is_multiple && (!@hide_results_on_select || (evt.metaKey or evt.ctrlKey))
-        this.winnow_results()
+        this.winnow_results(skip_highlight: true)
       else
         this.results_hide()
         this.show_search_field_default()
--- a/coffee/chosen.jquery.coffee
+++ b/coffee/chosen.jquery.coffee
@@ -366,7 +366,6 @@ class Chosen extends AbstractChosen
 
       @form_field.options[item.options_index].selected = true
       @selected_option_count = null
-      @search_field.val("")
 
       if @is_multiple
         this.choice_build item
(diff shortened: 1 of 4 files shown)


tjschuck:
> I spent… too much time (3:04)

Thanks, [Harvest](https://www.getharvest.com)! 😜

PR #2924, opened 2017-11-27
Make hide_results_on_select option apply for deselecting as well

VanTanev:
### Summary

`hide_results_on_select: false` option currently applies only for selecting options. It should apply on deselect as well, as custom skinning of chosen can make use of it.

Please double-check that:

  - [ ] If necessary, you've updated [the documentation](https://github.com/harvesthq/chosen/blob/master/public/options.html). -
  _I am unsure if this neccesitates a documentation change, as I think that this should have been the intended behavior of the option._


cgatesman:
I believe this request is the same as what I'm looking for.  To further explain, when using `hide_results_on_select: false`, it would be nice to be able to click on already selected items (which are currently greyed out) in order to unselect them (as if you clicked on the `x` of the `.search-choice`).

VanTanev:
Updated to also not scroll the dropdown menu to top when `hide_results_on_select: false`.

PR #2943, opened 2018-01-26
Remove forgotten usage of css `clip`.

### LINKED ISSUE #2816: Remove CSS clip property

### LINKED ISSUE #2942: Follow-up: Remove `clip` CSS property
@harvesthq/chosen-developers, particularly @adunkman: this is a follow up to PR #2939, which was to fix issue #2816.

Per [this comment on #2939](https://github.com/harvesthq/chosen/pull/2939#issuecomment-360336277), there's one `clip` remaining [here](https://github.com/harvesthq/chosen/blob/f86fa9b7f70f71a84971e6e4c2b48972d3db9aee/sass/chosen.scss#L146).

adunkman:
@harvesthq/chosen-developers to fix #2942 originally reported by @taichunmin, a follow-up to #2939 which fixed #2816.

I missed a usage of `clip` — when the search input is hidden due to search being disabled in options or having fewer than `disable_search_threshold` items. 

This usage of `clip` was similar to the others — to hide the search input but allow it to still receive keyboard events. We can use the same approach as #2939 — simply make the search input transparent instead.

### DIFF
--- a/sass/chosen.scss
+++ b/sass/chosen.scss
@@ -143,7 +143,8 @@ $chosen-sprite-retina: url('chosen-sprite@2x.png') !default;
   }
   &.chosen-container-single-nosearch .chosen-search {
     position: absolute;
-    clip: rect(0,0,0,0);
+    opacity: 0;
+    pointer-events: none;
   }
 }
 /* @end */


taichunmin:
LGTM

westonganger:
This causes problems with IE11 (I read somewhere pointer-events has issues in IE11.).  That being said its still better than the previous implementation.

After selecting an item the search input cursor line can still be seen overtop the select. 

Anyone know how I could fix this?

0nehand:
I have the same issue as @westonganger above.  After 1.8.3 was pushed, my selects no longer show what's actually selected.  Instead, the search input covers it up.  You can see it here:  http://alxlit.name/bootstrap-chosen/

Any word on a fix?

taichunmin:
@0nehand This is because bootstrap-chosen is outdated for 3 years. There are a lot of difference between chosen and bootstrap-chosen.

0nehand:
@taichunmin Ok thanks. I see there was an issue opened there and a workaround mentioned:
https://github.com/alxlit/bootstrap-chosen/issues/80

Issue #2953, opened 2018-02-13
IE 11 not working correctly with Chosen 1.8.3


wittdavid:
IE11 does not work correctly with keyboard actvitity with Chosen 1.8.3.  Prior versions seem to work okay.   

### Steps to reproduce

Tell us how to reproduce this issue.

  1. https://harvesthq.github.io/chosen/  Click in standard HTML dropdown in Standard Select.  TAB to chosen version of dropdown.

  2.  Note that browser shows the flashing input caret within the text of the placeholder.

  3. Press space bar, or down arrow, or press letter U.  All of these should cause the dropdown to appear.  Instead the dropdown flashes and disappears.

### Expected behavior

At step 2, browser should not show input caret.

At step 3, Dropdown should appear so that user can select, or input box in focus, as appropriate.

### Actual behavior

dropdown flashes and disappears.

### Environment

  - **Chosen Version**: 1.8.3

  - **jQuery or Prototype Version**:  n/a

  - **Browser and Version**:  IE 11

  - **OS and Version**:  Windows 7

### Additional information

Any other information you want to share that is relevant to the issue being reported. This might include the lines of code that you have identified as causing the bug, or potential solutions and workarounds.

bennybee:
Tested the above in Edge 14, 15 and 16 - they all show caret but step 3 works.

PR #2962, opened 2018-03-15
Use two inputs for Chosen’s focus and search.

adunkman:
@harvesthq/chosen-developers as a follow-up to #2939, fixing #2953.

As mentioned in #2939: 

> As an aside, I originally went with two inputs — a new one to catch the focus events when the results drop was closed, and the existing search input — however, it quickly became ridiculous because letters typed into the focus input cause the results drop to filter and show at the same time. This meant that I had to forward events that caused letters to be typed into the focus input over to the search input… and simulating keyboard events is gross.
> 
> Using a single input that moves ended up being significantly easier.

This isn’t my first choice, but IE11 has demonstrated interesting behavior when moving a focused input element in the DOM: 

- If an input is moved in the DOM during the processing of a `focus` event, it results in the input becoming blurred (this is the same cross-browser). 
- In IE11, the input never triggers a `blur` event, it just simply no longer is focused.
- In IE11, the browser seems to still considers the element focused, and therefore calling `focus()` on it does not cause it to become focused. 

Here’s [a minimal example](https://jsfiddle.net/y7ywv8hf/11/): 

| IE11 | Chrome | 
|-----|---------|
| screencast showing that blur events are not fired in IE11 | screencast showing blur events being fired in Chrome |

I dug into this quite a bit, but I wasn’t able to come up with a way to reliably move the input cross-browser.

The result: introducing a second input, registering the same event handlers on it, and then syncing the values of the inputs when jumping focus between the two inputs when the results drop is shown. It’s a little hacky, but I’m not confident that I can find a less-hacky solution. If you can think of one, I’m _so interested_ in doing that. 😅

### DIFF
--- a/coffee/chosen.jquery.coffee
+++ b/coffee/chosen.jquery.coffee
@@ -50,7 +50,8 @@ class Chosen extends AbstractChosen
     @form_field_jq.hide().after @container
     @dropdown = @container.find('div.chosen-drop').first()
 
-    @search_field = @container.find('input').first()
+    @search_field = @container.find('input.chosen-search-input')
+    @focus_field = @container.find('input.chosen-focus-input')
     @search_results = @container.find('ul.chosen-results').first()
     this.search_field_scale()
 
@@ -104,6 +105,18 @@ class Chosen extends AbstractChosen
       @search_choices.on 'click.chosen', (evt) => this.choices_click(evt); return
     else
       @container.on 'click.chosen', (evt) -> evt.preventDefault(); return # gobble click of anchor
+
+      @focus_field.on 'blur.chosen', (evt) => this.input_blur(evt); return
+      @focus_field.on 'focus.chosen', (evt) => this.input_focus(evt); return
+
+      transfer_value = () =>
+        @search_field.val(@focus_field.val())
+        @focus_field.val('')
+
+      @focus_field.on 'keyup.chosen', (evt) => transfer_value(); this.keyup_checker(evt); return
+      @focus_field.on 'keydown.chosen', (evt) => transfer_value(); this.keydown_checker(evt); return
+      @focus_field.on 'cut.chosen', (evt) => setTimeout(transfer_value, 0); this.clipboard_event_checker(evt); return
+      @focus_field.on 'paste.chosen', (evt) => setTimeout(transfer_value, 0); this.clipboard_event_checker(evt); return
 
   destroy: ->
     $(@container[0].ownerDocument).off 'click.chosen', @click_test_action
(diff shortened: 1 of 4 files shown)


coffee/chosen.jquery.coffee
```diff
@@ -179,10 +187,8 @@ class Chosen extends AbstractChosen
     @container.addClass "chosen-container-active"
     @active_field = true
 
-    @search_field.val(@search_field.val())
```
adunkman: This was actually a difference between the jQuery and Prototype versions — the Prototype version seen below is also removed, but it was: 

```coffeescript
@search_field.value = this.get_search_field_value()
```

That’s _almost_ the same, but not quite — resulting in the initial keypress being doubled in the Prototype version due to the change in that method (for example, keyboard focusing the element and pressing <kbd>R</kbd> resulted in `RR` in Prototype but `R` in jQuery). 

I couldn’t see a reason for setting the value of the field to… itself… so I dropped it completely to make the behavior consistent. 

Looking at the git history, it appears to originate in [465eb61](https://github.com/harvesthq/chosen/blame/465eb61eb690d01d2495e33a2b85f1000a0f1f6d/coffee/chosen.jquery.coffee#L182), the initial commit by @pfiller, humorously mentioning “Plenty of bugs at this point!”

sass/chosen.scss
```diff
@@ -70,6 +70,7 @@ $chosen-sprite-retina: url('chosen-sprite@2x.png') !default;
       cursor: pointer;
       opacity: 0;
       position: absolute;
+      width: 0;
```
adunkman: This small CSS change causes the carat to disappear in IE11. Why does the carat appear on-screen for a transparent input in the first place? I don’t have a good answer to that.

coffee/chosen.jquery.coffee
```diff
@@ -105,6 +106,18 @@ class Chosen extends AbstractChosen
     else
       @container.on 'click.chosen', (evt) -> evt.preventDefault(); return # gobble click of anchor
 
+      @focus_field.on 'blur.chosen', (evt) => this.input_blur(evt); return
+      @focus_field.on 'focus.chosen', (evt) => this.input_focus(evt); return
+
+      transfer_value = () =>
+        @search_field.val(@search_field.val() + @focus_field.val())
```
satchmorun: I left this concatenation in because @adunkman said there's a way to leave a single select and come back to it with a value present. I haven't been able to reproduce that, but I trust @adunkman.

adunkman: You may be trusting me more than I trust me… I’ll do some extra testing and try to figure it out, otherwise I’ll add a commit to drop the concat. 🐈

adunkman: Couldn’t figure it out! Must have been mistaken. Removed the concat and squashed the commits to prep for final review.

adunkman:
> I don't love having to do this…

Samesies. Thanks for the reviews and assistance, all! :heart:

PR #2988, opened 2018-05-27
resolved Composer

landsman:
fixed pull request in #2803


.gitignore
```diff
@@ -1,6 +1,7 @@
 .DS_Store
 node_modules
 .project
+.idea
```
koenpunt: Unnecessary change, should be in your global gitignore file IMO

composer.json
```diff
@@ -0,0 +1,36 @@
...
+    "input",
+    "ui"
+  ],
+  "type": "library",
+  "license": "MIT",
+  "minimum-stability": "stable",
+  "homepage": "https://harvesthq.github.io/chosen/",
+  "require": {
+  }
+}
```
koenpunt: Please add a trailing newline

README.md
```diff
@@ -28,6 +28,12 @@ To install with npm:
 npm install chosen-js
 ```
 
+To install with composer:
+
+```
+composer require harvesthq/chosen
```
koenpunt: For this to work it should still be added to packagist right?

composer.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "harvesthq/chosen",
```
koenpunt: To be sure; can the name be an arbitrary value? Because this won't match with the repo path `harvesthq/chosen-package`.

.gitignore
landsman: can we just let it here for others contributors?

README.md
landsman: Sure, it will after merge.

composer.json
landsman: Yeah, I think that can be.

.gitignore
koenpunt: Well since this is specific to a certain IDE I say no, because everyone can decide whatever IDE they use for js development. Was this an Android or iOS project I would agree in adding it to the projects gitignore file, because the options there are limited.

koenpunt: See also; https://intellij-support.jetbrains.com/hc/en-us/community/posts/115000154070-How-to-gitignore-idea-files?page=1#community_comment_115000258150

landsman: removed

composer.json
landsman: done

landsman:
Is it good right now?

landsman:
is it ready to merge please?

README.md
```diff
@@ -28,6 +28,12 @@ To install with npm:
 npm install chosen-js
 ```
 
+To install with composer:
```
tjschuck: Can you please capitalize `composer` and link it to https://getcomposer.org?

tjschuck:
I have created a `harvesthq` account on https://packagist.org in preparation.

tjschuck:
@stof @koenpunt Before we continue too much further, we were all pretty ¯\\\_(ツ)_/¯ about this on https://github.com/harvesthq/chosen/pull/2803 — do either of you actually have any arguments for or against including this particularly?

I've never heard of Composer/Packagist before, but I haven't written PHP in eons, so that doesn't mean much.  It does seem like it's the de facto PHP package manager, though.

That said, Chosen isn't a PHP library.  There's also stuff like https://asset-packagist.org that seems to be intended as a bridge between Composer and NPM/bower... is that a _better_ recommendation, or is that not as popular/reliable/trustworthy of a tool?

marclaporte:
@tjschuck : https://github.com/harvesthq/chosen/pull/2803#issuecomment-294078593 has several examples of JavaScript libs.

We (Tiki Wiki CMS Groupware) use https://asset-packagist.org/ and it's clearly a second choice only when we can't get from Packagist. 

I highly recommend to proceed with Packagist:
* It's a one-time effort, and after, it's all automated
* PHP developers look on Packagist for solutions. Most won't look on asset-packagist.org
* asset-packagist.org would be an extra hurdle

Thanks!

README.md
landsman: done

koenpunt:
I used Composer in my PHP days, and there's really no alternative in dependency management. And since it's automated (it uses tagged releases), I don't think this will hurt us.

landsman:
Exactly. Is full-automated solution for PHP world. For example of use-case is port to Drupal [Chosen module](https://www.drupal.org/project/chosen) which need manually install original library to `/libraries/` folder. It's annoying if you are use to on automated dependency workflow defined by one json file. After this merge it will be automated for CI/CD.

tjschuck:
Done! https://packagist.org/packages/harvesthq/chosen

Thanks again, everyone.

landsman:
Thanks @tjschuck !

JTubex:
@hussainweb opened a PR that relates to this one: #3066

PR #2989, opened 2018-05-29
Revert previous clip-deprecation fixes in favor of hybrid clip and clip-path.

adunkman:
@harvesthq/chosen-developers /cc @jisraelsen 

This PR reverts #2939 and #2943, which attempted to avoid using the deprecated `clip` property and used `display: none`, at the cost of a second input to capture focus events. 

It didn’t work out too well, and has continued to be a headache as we try to use the latest version in Harvest. I think the best thing to do is admit it was the wrong choice and revert it. Juggling focus and key events isn’t standard across browsers, especially while making DOM modifications.

This PR instead leaves the deprecated `clip` property and additionally introduces the same fix using `clip-path`. The result is that browsers that do not support the new standard `clip-path` will fallback to using the deprecated `clip` property, without requiring any additional changes.

----------------

One nice side-effect: while trying to make this all work correctly, I had to dig way into the way we handle focus and blur events in Chosen. If you’re not aware, we’re currently doing a lot of juggling around blur events, since blurring an input technically happens while clicking on an item to select it. I discovered that we can avoid doing all of that if we rely on `focusin` and `focusout` events on the wrapping container, as demonstrated [in this jsfiddle](https://jsfiddle.net/a6wgLjkn/15/) (watch for `chosenfocus` and `chosenblur` events). I’ll be pursuing this further in a separate PR, and expect to be able to delete a bunch of focus-specific properties as well as [the extra 100ms the chosen drop is visible](https://github.com/harvesthq/chosen/blob/5e11af9d432a73d4da08154c41b20456b3630e25/coffee/lib/abstract-chosen.coffee#L67-L70) — but didn’t want to hold this up on an unrelated change.

### DIFF
--- a/coffee/chosen.jquery.coffee
+++ b/coffee/chosen.jquery.coffee
@@ -243,9 +243,6 @@ class Chosen extends AbstractChosen
       @form_field_jq.trigger("chosen:maxselected", {chosen: this})
       return false
 
-    unless @is_multiple
-      @search_container.append @search_field
-
     @container.addClass "chosen-with-drop"
     @results_showing = true
 
@@ -261,10 +258,6 @@ class Chosen extends AbstractChosen
   results_hide: ->
     if @results_showing
       this.result_clear_highlight()
-
-      unless @is_multiple
-        @selected_item.prepend @search_field
-        @search_field.focus()
 
       @container.removeClass "chosen-with-drop"
       @form_field_jq.trigger("chosen:hiding_dropdown", {chosen: this})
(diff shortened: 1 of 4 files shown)


stof:
Where is the code dealing with focusin and focusout ? I don't see any added code relying on it.

koenpunt:
> Where is the code dealing with focusin and focusout ? I don't see any added code relying on it.
> Or did you mean this is a future change ?

It's not in here, according to the last sentence: 

> I’ll be pursuing this further in a separate PR, and expect to be able to delete a bunch of focus-specific properties as well as the extra 100ms the chosen drop is visible — but didn’t want to hold this up on an unrelated change.

adunkman:
Correct! I’ll be opening a follow-up PR today. Thanks for the review!

PR #2990, opened 2018-05-31
Remove 100ms lag on blur by using custom blur and focus events.

adunkman:
@harvesthq/chosen-developers /cc @jisraelsen 

This PR changes the way focus tracking occurs, relying on `focusin` and `focusout` events on the Chosen container instead of `focus` and `blur` events on the search input itself. The key benefit is that when the search input is blurred in the process of clicking on a search result item, focus is still maintained within the Chosen container. This allows us to drop our `mouse_on_container` tracking as well as the 100ms-delayed `blur_test`.

For a quick review of the events used here:

|              | `focus` | `focusin` | `blur` | `focusout` |
|---------|--------|-----------|--------|-----------|
| When? | When an element is focused. | When an element is focused. | When an element is blurred. | When an element is blurred. |
| Where? | Triggered on the element which is focused and does not bubble up the DOM. | Triggered on the element which is focused and every element which contains it (it bubbles up the DOM). | Triggered on the element which is focused and does not bubble up the DOM. | Triggered on the element which is focused and every element which contains it (it bubbles up the DOM). |

I’ll drop a few inline notes to explain more!

### DIFF
--- a/coffee/chosen.jquery.coffee
+++ b/coffee/chosen.jquery.coffee
@@ -34,6 +34,7 @@ class Chosen extends AbstractChosen
     container_props =
       'class': container_classes.join ' '
       'title': @form_field.title
+      'tabIndex': '-1'
 
     container_props.id = @form_field.id.replace(/[^\w]/g, '_') + "_chosen" if @form_field.id.length
 
@@ -78,6 +79,10 @@ class Chosen extends AbstractChosen
     @container.on 'mouseup.chosen', (evt) => this.container_mouseup(evt); return
     @container.on 'mouseenter.chosen', (evt) => this.mouse_enter(evt); return
     @container.on 'mouseleave.chosen', (evt) => this.mouse_leave(evt); return
+    @container.on 'focusin.chosen', (evt) => this.container_focusin(evt); return
+    @container.on 'focusout.chosen', (evt) => this.container_focusout(evt); return
+    @container.on 'chosen:blur.chosen', (evt) => this.input_blur(evt); return
+    @container.on 'chosen:focus.chosen', (evt) => this.input_focus(evt); return
 
     @search_results.on 'mouseup.chosen', (evt) => this.search_results_mouseup(evt); return
     @search_results.on 'mouseover.chosen', (evt) => this.search_results_mouseover(evt); return
@@ -93,10 +98,8 @@ class Chosen extends AbstractChosen
     @form_field_jq.on "chosen:open.chosen", (evt) => this.container_mousedown(evt); return
     @form_field_jq.on "chosen:close.chosen", (evt) => this.close_field(evt); return
 
-    @search_field.on 'blur.chosen', (evt) => this.input_blur(evt); return
     @search_field.on 'keyup.chosen', (evt) => this.keyup_checker(evt); return
     @search_field.on 'keydown.chosen', (evt) => this.keydown_checker(evt); return
-    @search_field.on 'focus.chosen', (evt) => this.input_focus(evt); return
     @search_field.on 'cut.chosen', (evt) => this.clipboard_event_checker(evt); return
     @search_field.on 'paste.chosen', (evt) => this.clipboard_event_checker(evt); return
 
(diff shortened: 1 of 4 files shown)


coffee/chosen.jquery.coffee
```diff
@@ -34,6 +34,7 @@ class Chosen extends AbstractChosen
     container_props =
       'class': container_classes.join ' '
       'title': @form_field.title
+      'tabIndex': '-1'
```
adunkman: When attempting to click on an item in the Chosen drop but missing by a pixel or two, we don’t want the Chosen drop to close. We can add a tab index, which allows it to receive focus — resulting in the following: 

0. The search input has focus, and the user attempts to use their mouse to click on a result.
1. The user begins clicking. The search input blurs, triggering a `focusout` event, which is caught on the container and a `setTimeout` is scheduled. 
2. The container receives the mouse event, which focuses the container. The container triggers a `focusin` event, which is caught on the container and nothing happens (`@active_field` is true).
3. The `setTimeout` expires, checking to see if the container contains `document.activeElement`, which is in this case, itself. As [defined in the spec](https://developer.mozilla.org/en-US/docs/Web/API/Node/contains), a Node `.contains()` itself, and therefore the drop is not closed.

coffee/chosen.jquery.coffee
```diff
@@ -76,8 +77,10 @@ class Chosen extends AbstractChosen
 
     @container.on 'mousedown.chosen', (evt) => this.container_mousedown(evt); return
     @container.on 'mouseup.chosen', (evt) => this.container_mouseup(evt); return
-    @container.on 'mouseenter.chosen', (evt) => this.mouse_enter(evt); return
-    @container.on 'mouseleave.chosen', (evt) => this.mouse_leave(evt); return
+    @container.on 'focusin.chosen', (evt) => this.container_focusin(evt); return
+    @container.on 'focusout.chosen', (evt) => this.container_focusout(evt); return
+    @container.on 'chosen:blur.chosen', (evt) => this.close_field(evt); return
+    @container.on 'chosen:focus.chosen', (evt) => this.input_focus(evt); return
```
adunkman: I think relying on custom events here (`chosen:blur` and `chosen:focus`) are the most straightforward way to handle this — but we can also just simply inline the calls to `input_focus` and `close_field` in the `container_focusin` and `container_focusout` events below. Let me know which you prefer!

sass/chosen.scss
```diff
@@ -8,6 +8,9 @@ $chosen-sprite-retina: url('chosen-sprite@2x.png') !default;
   vertical-align: middle;
   font-size: 13px;
   user-select: none;
+  &:focus {
+    outline: none;
```
tjschuck: @adunkman Just to double check: this just removes some extra extraneous focus on some unexpected element, right?  Chosen currently styles properly on focus like you'd expect (good for accessibility!), so I'm hoping this is just that!

coffee/lib/abstract-chosen.coffee
```diff
@@ -55,20 +54,12 @@ class AbstractChosen
...
     if @is_multiple
       setTimeout (=> this.container_mousedown()), 50 unless @active_field
     else
       @activate_field() unless @active_field
 
-  input_blur: (evt) ->
-    if not @mouse_on_container
-      @active_field = false
-      setTimeout (=> this.blur_test()), 100
-
```
adunkman: Since we now only trigger our `chosen:focus` and `chosen:blur` events when focus is lost on the entire container or its descendants, we no longer need to track mouse positioning.

sass/chosen.scss
adunkman: Correct! This removes a focus rectangle around the Chosen DIV itself, which only occurs when clicking on search results (after the search input has blurred but before the search result is selected). It’s quite jarring, since it’s an unexpected element to receive focus in this case.

koenpunt:
Looks promising, but does is work cross browser?

coffee/chosen.jquery.coffee
koenpunt: This suggests that those events can/should be invoked from a client implementation, which I think is not the case, right? So I think inlining them below would be better.

PR #2998, opened 2018-06-08
Add docs about how to release Chosen versions

tjschuck:
@harvesthq/chosen-developers @adunkman Let's document this!  I'm like... 90% sure this is right.  It's at least how I released the most recent version just now!

Rendered version viewable here: https://github.com/harvesthq/chosen/blob/e7ec809b911da27464cac57a0d6bb2f63df068ac/contributing.md#release-a-new-version-of-chosen

### DIFF
--- a/contributing.md
+++ b/contributing.md
@@ -191,3 +191,29 @@ Once you're configured, `grunt` tasks are available:
     grunt watch                # watch coffee/ for changes and build Chosen
 
 If you're interested, you can find the task in [Gruntfile.coffee](https://github.com/harvesthq/chosen/blob/master/Gruntfile.coffee).
+
+# Release a new version of Chosen
+
+This documentation is for Chosen maintainers.  You must have write permissions for this repository to cut a release.
+
+1. Draft a new release [on the releases page](https://github.com/harvesthq/chosen/releases).
+
+    1. Title formatted as "Version X.Y.Z"
+    2. Tagged as `vX.Y.Z`
+    3. Body of the release should contain changes included in the release. Ideally, there will already be a draft release present with a running log from contributions merged since the last release.
+
+2. Bump the version number in the source. [Example commit](https://github.com/harvesthq/chosen/commit/be0a298f528ec59ce97889eaeeeb47a2dca9ca79).
+
+3. Run `grunt prep-release`
+
+4. Attach the generated `chosen_vX.Y.Z.zip` file (in the Chosen root directory) to the draft release from step 1.
+
+5. Target the release tag to the commit from step 2.
+
+6. Publish the release.
+
+7. Run `grunt publish-release` to publish to the `gh-pages` branch.
+
+8. Verify https://harvesthq.github.io/chosen/ is showing the new version.
+
+_Note:_ the release repository [harvesthq/chosen-package](https://github.com/harvesthq/chosen-package) is automatically built by [Travis CI](https://travis-ci.org/harvesthq/chosen). No explicit action is necessary.
--- a/contributing.md
+++ b/contributing.md
@@ -196,24 +196,29 @@ If you're interested, you can find the task in [Gruntfile.coffee](https://github
 
 This documentation is for Chosen maintainers.  You must have write permissions for this repository to cut a release.
 
-1. Draft a new release [on the releases page](https://github.com/harvesthq/chosen/releases).
+1. Bump the version number in the source but _do not push it yet_. [Example commit](https://github.com/harvesthq/chosen/commit/be0a298f528ec59ce97889eaeeeb47a2dca9ca79).
 
-    1. Title formatted as "Version X.Y.Z"
-    2. Tagged as `vX.Y.Z`
-    3. Body of the release should contain changes included in the release. Ideally, there will already be a draft release present with a running log from contributions merged since the last release.
+2. Create a tag for that commit: `git tag <tagname> <SHA>`
 
-2. Bump the version number in the source. [Example commit](https://github.com/harvesthq/chosen/commit/be0a298f528ec59ce97889eaeeeb47a2dca9ca79).
+    - `<tagname>` formatted as `vX.Y.Z`
+    - `<SHA>` is the SHA of the commit from step 1
 
-3. Run `grunt prep-release`
+3. Push the commit _and_ the tag: `git push origin --follow-tags`
 
-4. Attach the generated `chosen_vX.Y.Z.zip` file (in the Chosen root directory) to the draft release from step 1.
+4. Draft a new release [on the releases page](https://github.com/harvesthq/chosen/releases).
 
-5. Target the release tag to the commit from step 2.
+    - Title formatted as "Version X.Y.Z"
+    - Tagged as `<tagname>` from step 2
+    - Body of the release should contain changes included in the release. Ideally, there will already be a draft release present with a running log from contributions merged since the last release.
 
-6. Publish the release.
+5. Run `grunt prep-release`
 
-7. Run `grunt publish-release` to publish to the `gh-pages` branch.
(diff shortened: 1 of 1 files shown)


tjschuck:
Actually, there's an order issue here... the automatic build on Travis is dependent on the tag existing.

So... it seems like it'd be better to make the "bump the version" commit first, tag that explicitly, and then use the existing tag for the release.

Updating!  Gimme a few...

tjschuck:
Okay, 4407fa1 should fix that issue.

New rendered version here: https://github.com/harvesthq/chosen/blob/4407fa143dc1e9cf449cdc54d7c5277178a95239/contributing.md#release-a-new-version-of-chosen

koenpunt:
Should this be in contributing? Maybe a separate file (e.g.`deployment.md`) is better?

contributing.md
```diff
@@ -191,3 +191,34 @@ Once you're configured, `grunt` tasks are available:
     grunt watch                # watch coffee/ for changes and build Chosen
 
 If you're interested, you can find the task in [Gruntfile.coffee](https://github.com/harvesthq/chosen/blob/master/Gruntfile.coffee).
+
+# Release a new version of Chosen
+
+This documentation is for Chosen maintainers.  You must have write permissions for this repository to cut a release.
+
+1. Bump the version number in the source but _do not push it yet_. [Example commit](https://github.com/harvesthq/chosen/commit/be0a298f528ec59ce97889eaeeeb47a2dca9ca79).
```
koenpunt: Why the "do not push it yet"? You can create tags for already pushed commits, without changing history.

koenpunt: After inspecting the process, the steps are actually; update `package.json`, run `grunt prep-release` to update the version in the generated files, and only after that a commit should be made.

koenpunt: We could actually add the commit and tag steps to the `prep-release` task

tjschuck: @koenpunt The "do not push yet" was part of the first issue I commented upon — the Travis build step is:

1. Triggered by a commit hitting master
2. Dependent on the version tag existing

So if you pushed the commit first, the build would happen without the tag and the version would be wrong.

---

As for "update `package.json`, run `grunt prep-release`", that misses the [version string in `package-lock.json`](https://github.com/harvesthq/chosen/blob/be0a298f528ec59ce97889eaeeeb47a2dca9ca79/package-lock.json#L3).

It looks like committing package-lock.json is a recent change: https://github.com/harvesthq/chosen/commit/b0ac854ea0d3771bf3af2796d193bafb25770aef#diff-32607347f8126e6534ebc7ebaec4853d — should we be doing that?

tjschuck:
> Should this be in contributing? Maybe a separate file (e.g. `deployment.md`) is better?

I thought about that!  I went with this because I thought "what would be most likely to be actually found by future-me?"  That said, the answer wasn't super cut-and-dry, so I'd be willing to move it (maybe to `RELEASING.md`? [Here's the only other example I know of offhand](https://github.com/rails/rails/blob/c18621fd7c2f912aa6fca214d045ecfc303ae37f/RELEASING_RAILS.md), although I'm sure there are countless).

Any thoughts/preferences among @harvesthq/chosen-developers ?

adunkman:
Thanks for figuring this out, @tjschuck! I never really got it right, I just fumbled through it and had to run some steps locally to get it working each time I pushed a new version. 

If I were improving this in the future, I’d drop all the automated steps on CI, and instead opt for everything in a grunt task to be run locally. The reason for this is that we can no longer use CI to publish automatically, since the chosen NPM account now requires a two-factor auth code — so if we already have to run a manual process which authorizes as chosen’s NPM account, we might as well just put it all together!

tjschuck:
> The reason for this is that we can no longer use CI to publish automatically, since the chosen NPM account now requires a two-factor auth code

Cool, so I didn't actually release the new versions to NPM 😓 

I guess I need to fix that too and document whatever those steps are here!  I'll try to do that soon...

adunkman:
You can verify at https://www.npmjs.com/package/chosen-js — looks like the latest published version is 1.8.5.

tjschuck:
Based on [the age of this line](https://github.com/harvesthq/chosen/blob/be0a298f528ec59ce97889eaeeeb47a2dca9ca79/package-travis.yml#L12) (added in https://github.com/harvesthq/chosen/pull/2550 in June 2016), and the age of the token listed at https://www.npmjs.com/settings/chosenjs/tokens ("Created 2 months ago"), I'm guessing someone rotated that token without knowing what it was used for...

I'm going to regenerate the token in NPM _and_ update the encrypted version in Chosen to see if that fixes things, _and_ add docs for that token as well!

tjschuck:
Oh, which is to say: I don't think 2FA has anything to do with this.  We should still be able to use an API token.

contributing.md
koenpunt: But when there's no new tag, the publish script just doesn't do anything, and when then the tag is pushed, the script runs again and publishes the release. Or at least I'm pretty sure I've always released it like that. But I might of course be wrong. 
In that case I think you should add your comment from above to the document too, to clarify why one should not push yet

---

Regarding `package-lock.json`; it can be safely added. 

Updating the lock file is as easy as running `npm install` I think.

But the other day I have been looking at using `npm version`, for automating updating the version in the `package.json`, and can be configured with some hooks to automatically run `grunt prep-release` too.

I'll see if I can have that working today.

tjschuck:
#3003 is merged (fixing the token for automatic NPM releases via Travis), but that will only work on future releases, since it's only triggered for tagged commits.

I'm going to manually upload the missing versions to NPM now (1.8.6 and 1.8.7).  Stay tuned!

tjschuck:
Uhhh, weirdly, it DID try to upload the latest version, because I guess the tags never got pushed to the chosen-package repo last time.  You can see the triggering of `publish-package.sh` here: https://travis-ci.org/harvesthq/chosen/jobs/397973866#L1193

Which, fine, but then that also broke!  Check out the build for `chosen-package` here, which fails because `package-lock.json` ends up dirty, which kills the build: https://travis-ci.org/harvesthq/chosen-package/builds/397974855#L514

I'm looking into that now!

tjschuck:
The dirty `package-lock.json` was a red herring!  The real error message was hidden behind a collapsed log — it's here: https://travis-ci.org/harvesthq/chosen-package/builds/397974855#L509

Working on fixing that now!

tjschuck:
Okay!  Automatic publishing to NPM should now be fixed.

After all that, with failed builds left and right, we eventually got back to a known failure, and guess what: @adunkman was right all along!  There was a 2FA problem.  It turns out that NPM requires 2FA even when you're using API tokens, like... no one else ever.  Thankfully (but strangely still), they have two "levels" of 2FA — an "authentication only" one for logging in to the website and doing stuff with the CLI, and an "authentication + publish" level, that is all the above + requires the 2FA code for publishing.  We just "downgraded" to the "authentication only" one, and now everything works!  Sorry, @adunkman 🙃  See also: bbbaa1a92c6cde53092a98bc17b2bf38d92216e2, but that commit message is mostly just the same information.

I also further clarified the docs on how to encrypt the key here: https://github.com/harvesthq/chosen/blob/7149a2aed86d4d815f1497f27ae602ed898b9b02/package-travis.yml#L12-L20  Specifically, that `-r "harvesthq/chosen-package"` argument is necessary, because even though the token is canonically stored in the main Chosen repo, it's _used_ from the chosen-package repo.

I've also now pushed all the missing versions up to NPM.

More coming back about _this_ PR in a bit!

tjschuck:
@satchmorun and I were having difficulties with `git push origin --follow-tags` not actually pushing tags like we were wanting it to, and careful reading of the docs and countless Googles eventually made me realize it only works with _annotated_ tags.  Updated these docs in 4133dd9 to use the tag name also as the message to force the tag to be annotated so that it would be pushed at once.

PR #3003, opened 2018-06-28
Update encrypted NPM release token and document it

tjschuck:
@harvesthq/chosen-developers See [this comment](https://github.com/harvesthq/chosen/pull/2998#issuecomment-396400137) and the ~3 following ones.

This token has been broken for a while, so I generated a new one, encrypted it, updated the encrypted version, and added documenting comments about where this all comes from for the future.

### DIFF
--- a/package-travis.yml
+++ b/package-travis.yml
@@ -9,7 +9,11 @@ deploy:
   provider: npm
   email: chosenjs@getharvest.com
   api_key:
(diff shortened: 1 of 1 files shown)


tjschuck:
Gonna go ahead and merge this — this probably won't actually trigger a build + upload the old versions missing from NPM (currently 1.8.6 and 1.8.7) because it [only runs on tagged commits](https://github.com/harvesthq/chosen/blob/be0a298f528ec59ce97889eaeeeb47a2dca9ca79/package-travis.yml#L14).

I'll manually upload those releases shortly, but will talk about it on #2998 instead.  Consider this PR dead to me.  #2998 is my true favorite.

PR #3004, opened 2018-06-28
Don't run a matrix against node latest; just the version we support (6)

tjschuck:
@satchmorun and I discovered this while pairing on the automatic build stuff — though `node` was failing in the matrix (which is fine), if it ever _succeeded_, either accidentally or intentionally, it would run the publish script twice — once for each success.

Instead of testing against something we know is going to fail without a goal of fixing it, let's just remove it until we think we're going to make it pass.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ language: node_js
 
 node_js:
   - 6
-  - node
 
 addons:
   apt:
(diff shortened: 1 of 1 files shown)


tjschuck:
This passed on CI, but no hooks back to Github?  Weird.

https://travis-ci.org/harvesthq/chosen/builds/398027827

Issue #3005, opened 2018-07-07
prism.js - outdated, source-less binary


onlyjob:
[prism.js](https://github.com/harvesthq/chosen/blob/master/public/docsupport/prism.js) hasn't been touched since 2013 even though it is semantically versioned, [well maintained library](https://github.com/PrismJS/prism/releases).

On top of that committed version is lacking source which is a serious breach of open source compliance because effectively minified code is equivalent of pre-compiled binary -- un-readable, non-modifiable and non-distributable in [Debian](https://debian.org).  
It is a bad practice to commit binaries to VCS.

Please replace minified `prism.js` with original, uncompressed, up-to-date version.
I recommend using the following URL:

  https://github.com/PrismJS/prism/raw/v1.15.0/prism.js

which I would note into `prism.js.URL` file to document file's origin.

Read more about pre-minified files here https://wiki.debian.org/onlyjob/no-minification

Thanks.

koenpunt:
Not sure what you mean; but the prism.js file contains a license header, so I don't really see a problem there.

onlyjob:
Availability of source code is a foundation of [open source](https://opensource.org/osd-annotated).
Here _prism.js_ is provided only in compiled form, without source so license header is not good enough.

How do you know if _prism.js_ hasn't been tampered with? It is not possible to re-build this file from source because source is missing...

Also it would be nice to update it to current version. Please? ;)

koenpunt:
> Availability of source code is a foundation of open source.

True, and we provide the source code of Chosen. But since we're not the author of Prism, and we're not distributing Prism itself as software, I believe we are free to use a minified version, as long as we mention the author/refer to the license.

> It is not possible to re-build this file from source because source is missing...

Again, since we're not the author of Prism, there's no point in re-building the file. And if we want, we can always go to [the source](https://github.com/PrismJS/prism/), and create a new minified version.

> Also it would be nice to update it to current version. Please? ;)

Feel free to submit a PR! 😃

onlyjob:
No it is not that simple. It is true that license pf Prism allows you to re-distribute binaries.
But it is good practice to distribute all binaries with source.

Your source distribution is incomplete without Prism source and obtaining correct source for this particular binary is not possible: its version is unknown and the librrary is so old that it pre-dates first Priism release. You can get the different version of the source but not for this particular binary.
You wouldn't be able to reproduceibly re-build the exact minifified binary which is committed to the source tree here.

[...]

PR #3006, opened 2018-07-08
Updated "prism.js" to un-minified v1.15.0 (Closes: #3005)

onlyjob:
Open source compliance. Documented origin URL and LICENSE.

I'm unable to do full library transition.
Some CSS and code updates may be required to complete Prism update.

### DIFF
--- a/public/docsupport/prism.LICENSE
+++ b/public/docsupport/prism.LICENSE
@@ -1,0 +1,21 @@
+MIT LICENSE
+
+Copyright (c) 2012 Lea Verou
+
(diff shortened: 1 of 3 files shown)


koenpunt (changes requested):
I definitely think that we're not required to provide the .license and/or .url files.
However I would accept a PR that updates the (minified) Prism library, and the usage thereof.

onlyjob:
Ideally it would not be required to provide license and URL files yet it is a good idea (and good practice) to document origin of the file and its license.

On this instance LiICENSE and URL files are _required_ because file's header contains no information about its license, version or origin.

PR #3026, opened 2018-10-04
Colors moved to variables in SCSS

LorincJuraj:
This will simplify theming process and integration into existing project.

### DIFF
--- a/sass/chosen.scss
+++ b/sass/chosen.scss
@@ -1,6 +1,35 @@
 $chosen-sprite: url('chosen-sprite.png') !default;
 $chosen-sprite-retina: url('chosen-sprite@2x.png') !default;
 
+$bg-color: #fff;
+$box-shadow-color: #000;
+$border-color: #aaa;
+$group-name-color: #999;
+$highlited-color: #3875d7;
+$highlited-second-color: #2a62bc;
+
+$search-choice-bg-color: #eee;
+$search-choice-disabled-bg-color: #e4e4e4;
+$search-choice-color: #333;
+$search-choice-bg: linear-gradient(#f4f4f4 20%, #f0f0f0 50%, #e8e8e8 52%, #eee 100%);
+$search-choice-disabled-color: #666;
+$search-choice-focus-bg-color: $search-choice-disabled-color;
+$search-choice-disabled-border-color: #ccc;
+$no-results-bg: #f4f4f4;
+
+$chosen-color: #222;
+$chosen-border-color: #5897fb;
+$chosen-single-color: #444;
+$chosen-bg: linear-gradient($bg-color 20%, #f6f6f6 50%, #eee 52%, #f4f4f4 100%);
+$chosen-single-bg: linear-gradient(#eee 20%, $bg-color 80%);
+$chosen-default-color: $group-name-color;
+$chosen-container-color: $chosen-single-color;
+$chosen-container-disabled-color: $search-choice-disabled-border-color;
+$chosen-container-no-results-color: #777;
+$chosen-choices-color: #999;
+$chosen-choices-bg: linear-gradient($search-choice-bg-color 1%, $bg-color 15%);
+$chosen-drop-result-selected: $search-choice-disabled-border-color;
+
 /* @group Base */
 .chosen-container {
   position: relative;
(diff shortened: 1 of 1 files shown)


sass/chosen.scss
```diff
@@ -1,6 +1,35 @@
 $chosen-sprite: url('chosen-sprite.png') !default;
 $chosen-sprite-retina: url('chosen-sprite@2x.png') !default;
 
+$bg-color: #fff;
```
stof: All variable names should start with `chosen-` to avoid conflicts with other styles for people importing the Scss in their own files.

sass/chosen.scss
```diff
@@ -1,6 +1,35 @@
...
+$box-shadow-color: #000;
+$border-color: #aaa;
+$group-name-color: #999;
+$highlited-color: #3875d7;
+$highlited-second-color: #2a62bc;
```
stof: typo in the spelling of `highlighted`

sass/chosen.scss
```diff
@@ -1,6 +1,35 @@
 $chosen-sprite: url('chosen-sprite.png') !default;
 $chosen-sprite-retina: url('chosen-sprite@2x.png') !default;
 
+$chosen-bg-color: #fff;
+$chosen-box-shadow-color: #000;
+$chosen-border-color: #aaa;
+$chosen-group-name-color: #999;
+$chosen-highlighted-color: #3875d7;
+$chosen-highlighted-second-color: #2a62bc;
```
mlettini: I'm not a fan of using two different colors here to represent a gradient. You have gradients in variables below, so why not do the same here. Also, we could clarify that these are background colors. I'd recommend:

```
$chosen-highlighted-bg-color: #3875d7;
$chosen-highlighted-bg-gradient: linear-gradient(#3875d7 20%, #2a62bc 90%);
```

sass/chosen.scss
```diff
@@ -1,6 +1,35 @@
...
+$chosen-search-choice-disabled-bg-color: #e4e4e4;
+$chosen-search-choice-color: #333;
+$chosen-search-choice-bg: linear-gradient(#f4f4f4 20%, #f0f0f0 50%, #e8e8e8 52%, #eee 100%);
+$chosen-search-choice-disabled-color: #666;
+$chosen-search-choice-focus-bg-color: $chosen-search-choice-disabled-color;
+$chosen-search-choice-disabled-border-color: #ccc;
+$chosen-no-results-bg: #f4f4f4;
+
+$chosen-color: #222;
+$chosen-border-color: #5897fb;
```
mlettini: This is the first actual issue I see in the code. `$chosen-border-color` is already being defined above. Looking at the CSS, `#5897fb` is only used when active/focused. So I think this should be `$chosen-active-border-color`. I see some other active/focused-specific variables, and I think those variables should be grouped together, and have consistent naming.

sass/chosen.scss
```diff
@@ -1,6 +1,35 @@
...
+$chosen-single-color: #444;
```
mlettini: Same "text" comment here: `$chosen-single-text-color`

sass/chosen.scss
```diff
@@ -1,6 +1,35 @@
...
+$chosen-bg: linear-gradient($chosen-bg-color 20%, #f6f6f6 50%, #eee 52%, #f4f4f4 100%);
+$chosen-single-bg: linear-gradient(#eee 20%, $chosen-bg-color 80%);
+$chosen-default-color: $chosen-group-name-color;
+$chosen-container-color: $chosen-single-color;
+$chosen-container-disabled-color: $chosen-search-choice-disabled-border-color;
+$chosen-container-no-results-color: #777;
+$chosen-choices-color: #999;
+$chosen-choices-bg: linear-gradient($chosen-search-choice-bg-color 1%, $chosen-bg-color 15%);
+$chosen-drop-result-selected: $chosen-search-choice-disabled-border-color;
```
mlettini: I'm not sure how I feel about using a variable inside the definition of another variable. Normally I'd be all for it, but in this case (and a few other cases I see in this file), the variables are completely unrelated. For instance, `$chosen-drop-result-selected` has absolutely nothing to do with `$chosen-search-choice-disabled-border-color` (other than being a similar hex color), and by making the former defined by the latter, the code is making the two seem related in some way. I'll let another dev make a call on that (cc @stof)

mlettini (changes requested):
I love this idea of moving colors into variables to prepare for better theming. Nice work @LorincJuraj.

I didn't actually finish leaving all the comments I had, because I realized 1. I was adding those comments individually instead of all at once with a review, which can be annoying for notifications (sorry didn't mean to do that), and 2. I was repeating myself regarding naming conventions. Most of my comments are either about naming, or which variables we group together. So my feedback can be boiled down to essentially:

* Let's keep the naming consistent and visual. If the color is for a background color, let's use `-bg-color:`, for background gradient `-bg-gradient`, for text color `-text-color`, etc etc.
* Let's keep the order and grouping of variables consistent with how they are used in the CSS below. So all `-bg-color:`s and `-bg-gradients:` should be next to each other, all active/focus colors grouped together, disabled, single, multi, etc etc.

So take that into account with all the variables, including the lines I didn't comment on. ✌️

PR #3047, opened 2019-03-21
jQuery.fn.blur()/.focus() event shorthand is deprecated

aTei:
https://github.com/jquery/jquery-migrate/blob/master/warnings.md#jqmigrate-jqueryfnclick-event-shorthand-is-deprecated


aTei:
Should I close the PR?

PR #3082, opened 2020-01-05
Add support for chosen:search_updated event

kir:
The event is fired on the original selector when Chosen’s search results are updated after typing the search term.

The event also includes a corresponding `search_term` parameter.

### DIFF
--- a/coffee/chosen.jquery.coffee
+++ b/coffee/chosen.jquery.coffee
@@ -255,6 +255,9 @@ class Chosen extends AbstractChosen
   update_results_content: (content) ->
     @search_results.html content
 
+  fire_search_updated: (search_term) ->
+    @form_field_jq.trigger("chosen:search_updated", {chosen: this, search_term: search_term})
+
   results_hide: ->
     if @results_showing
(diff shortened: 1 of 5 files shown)


tjschuck:
@kir Can you give us a realistic use case for this event?

kir:
@tjschuck I'm using this event to update the list of items shown via selector, i.e. there is a delayed ajax request to the server which depends on the filter contents and I re-populate the selector with an alternative item set, depending on the filter.

kir:
Guys, any follow-up on this? I've used this patch in https://chrome.google.com/webstore/detail/checkvist-web-clipper/pheahjlflgbnajeilgglgcdlbkdckamf extension and it works nicely.

Unfortunately, I have troubles publishing the same extension in the Firefox add-on repository as they insist on using [non-patched version of the libraries](https://extensionworkshop.com/documentation/publish/third-party-library-usage).

So I'd really appreciate a release with my patch.

jedi-vim:
Any prediction to release this feature?

PR #3134, opened 2022-08-15
Issue #3133: Use placeholder attribute for placeholder text.

### LINKED ISSUE #3133: Use `placeholder` attribute on input elements for placeholder text (instead of `value` attribute)
Chosen puts the Placeholder text into the `value` attribute on the input elements, instead of in the `placeholder` attribute. It still renders as though it was a placeholder.  

I'm using the localize.js service for translation on several websites. This service specifically searches for placeholder attributes and translates those, so that all visible text on the page can be translated.  The localize script is not able to recognize the placeholder text, because it's not in the placeholder attribute, so these values are not properly translated.

Is there some reason the `placeholder` attribute was not used for its intended purpose? Does the `placeholder` attribute not work as expected?

I would like to propose that Chosen switch back to using attributes as they were intended to avoid this problem, and similar conflicts with other projects or systems that are expecting standard markup.

jenlampton:
Fixes https://github.com/harvesthq/chosen/issues/3133

### Summary

* Switched placeholder text from `value` attribute to `placeholder` attribute on `input` element.

Please double-check that:

  - [ ] You used [Grunt](https://github.com/harvesthq/chosen/blob/master/contributing.md#grunt) to build the JavaScript files and tested them locally.
  - [ ] You've updated both the jQuery *and* Prototype versions.
  - [x] N/A If necessary, you've updated [the documentation](https://github.com/harvesthq/chosen/blob/master/public/options.html).

### References

If your pull request is in reference to one or more open GitHub issues, please mention them here to keep the conversations linked together:
- https://github.com/harvesthq/chosen/issues/3133

### DIFF
--- a/coffee/lib/abstract-chosen.coffee
+++ b/coffee/lib/abstract-chosen.coffee
@@ -347,7 +347,7 @@ class AbstractChosen
     """
       <ul class="chosen-choices">
         <li class="search-field">
-          <input class="chosen-search-input" type="text" autocomplete="off" value="#{@default_text}" />
+          <input class="chosen-search-input" type="text" autocomplete="off" placeholder="#{@default_text}" />
         </li>
       </ul>
       <div class="chosen-drop">


jenlampton:
The command `npm install` failed with the error `The package-lock.json file was created with an old version of npm` and a handful of deprecations, so I was not able to use Grunt to build the JavaScript files.

To test locally, I modified the `chosen.jquery.js` file directly since the change was very simple. If someone else can get grunt working with the latest version of chosen, I could use some help getting the build process going for this PR.

